PR #19, opened 2016-08-17
Implement image URL form and display

### LINKED ISSUE #2: Save Image Link
#### As a user I want to save a link to an image.

__Story__:

Now that we have an awesome landing page for our users to experience it is time
to introduce our first feature! In general, we want to deliver value to our
customers as fast as possible. As a result, we have decided that we want our
users to be able to save links to images, and immediately see that we have
saved the URL to the image.

One could argue that simply saving the image, and not showing it, provides some
value. Yes, you could split this story here. However, because saving image URLs
and viewing those images are both relatively small additions that work well
together, it is acceptable to complete them together. Note, on the other hand,
that we do not yet want to implement the image index. Its addition adds too
much complexity, so we’ll save that for a future story.

In general, feature addition, or stories, should be split into the smallest
deliverable chunk that provides value to your users with the caveat that it
might make sense to combine two, sometimes more, tiny and related features into
a single story.

__Acceptance criteria__:
- [x] The landing page links to the image link submission form.
- [x] The link is entered through a form.
- [x] After the form is saved, the link is persisted in the database.
- [x] After the form is submitted, the user is redirected to a page displaying
  the image.

__Dependencies__:
- Deployment Environment

__Note__: Many image sharing sites will permit one to upload images, or
download images when provided a link. For this project we will only work with
the link, i.e., URL, to various images.

ragurney:
Closes #2

### DIFF
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -1,0 +1,5 @@
+class ImagesController < ApplicationController
+  def new
+    @image = Image.new
+  end
+end
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -2,4 +2,21 @@ class ImagesController < ApplicationController
   def new
     @image = Image.new
   end
+
+  def create
+    @image = Image.new(image_params)
+    if @image.save
+      flash[:success] = 'Url successfully saved!'
+      redirect_to image_path(id: @image.id)
+    else
+      flash[:error] = 'Url cannot be empty'
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def image_params
+    params.require(:image).permit(:url)
+  end
 end
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -14,6 +14,12 @@ class ImagesController < ApplicationController
     end
   end
 
+  def show
+    @image = Image.find(params['id'])
+  rescue ActiveRecord::RecordNotFound => _e
+    render plain: 'The image you were looking for was not found!', status: :not_found
+  end
+
   private
 
   def image_params
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -1,0 +1,3 @@
+class Image < ActiveRecord::Base
+  validates :url, presence: true
+end
--- a/app/views/images/new.html.erb
+++ b/app/views/images/new.html.erb
@@ -1,0 +1,12 @@
+<h1>New Image URL</h1>
+
+<%= form_for :image, url: images_path  do |f| %>
+  <p>
+    <%= f.label :url %><br>
+    <%= f.text_field :url %>
+  </p>
+
+  <p>
+    <%= f.submit %>
+  </p>
+<% end %>
--- a/app/views/images/show.html.erb
+++ b/app/views/images/show.html.erb
@@ -1,0 +1,1 @@
+<img src=<%= @image.url %> >
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -1,0 +1,10 @@
+require 'test_helper'
+
+class ImagesControllerTest < ActionDispatch::IntegrationTest
+  test 'should get images form page' do
+    get new_image_path
+
+    assert_response :ok
+    assert_select 'h1', 1, 'New Image URL'
+  end
+end
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -1,10 +1,29 @@
 require 'test_helper'
 
 class ImagesControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @url = 'http://jstsolutions.net/wp-content/themes/realty/lib/images/key_img2.png'
+    @image = Image.create(url: @url)
+  end
+
   test 'should get images form page' do
     get new_image_path
 
     assert_response :ok
     assert_select 'h1', 1, 'New Image URL'
+  end
+
+  test 'controller should create and put correctly' do
+    post images_path(image: { url: @url })
+
+    assert_response :found
+    assert_not_nil Image.find_by(url: @url)
+  end
+
+  test 'invalid url does not create image' do
+    post images_path(image: { url: nil })
+
+    assert_response :unprocessable_entity
+    assert_equal 'Url cannot be empty', flash[:error]
   end
 end
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -13,6 +13,20 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_select 'h1', 1, 'New Image URL'
   end
 
+  test 'should show correct image' do
+    get image_path(@image)
+
+    assert_response :ok
+    assert_select 'img', 1
+  end
+
+  test 'should show error for incorrect image' do
+    get image_path(id: -1)
+
+    assert_response :not_found
+    assert_equal 'The image you were looking for was not found!', response.body
+  end
+
   test 'controller should create and put correctly' do
     post images_path(image: { url: @url })
 
--- a/test/models/image_test.rb
+++ b/test/models/image_test.rb
@@ -1,0 +1,13 @@
+require 'test_helper'
+
+class ImageTest < ActiveSupport::TestCase
+  def test_url__invalid
+    img = Image.new
+    refute img.valid?
+  end
+
+  def test_url__valid
+    img = Image.new(url: 'http://placehold.it/120x120&text=image1')
+    assert img.valid?
+  end
+end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -9,6 +9,9 @@
   </head>
 
   <body>
+    <% flash.each do |key, value| %>
+      <div><%= value %></div>
+    <% end %>
     <%= yield %>
   </body>
 </html>
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -1,1 +1,2 @@
 <h1>Yay, a homepage!</h1>
+<%= link_to 'Add new image...', new_image_path %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :images, only: [:new, :create, :show]
+
   root 'pages#index'
 end
--- a/db/migrate/20160817175443_create_images.rb
+++ b/db/migrate/20160817175443_create_images.rb
@@ -1,0 +1,7 @@
+class CreateImages < ActiveRecord::Migration[5.0]
+  def change
+    create_table :images do |t|
+      t.string :url, null: false
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,6 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 0) do
+ActiveRecord::Schema.define(version: 20160817175443) do
+
+  create_table "images", force: :cascade do |t|
+    t.string "url", null: false
+  end
 
 end
--- a/test/flows/images_flow_test.rb
+++ b/test/flows/images_flow_test.rb
@@ -1,0 +1,27 @@
+require 'flow_test_helper'
+
+class ImagesFlowTest < FlowTestCase
+  setup do
+    @valid_url = 'http://digital-photography-school.com/wp-content/uploads/2007/02/black-and-white-tips.jpg'
+    @invalid_url = 'asdf https://oh_no.this/?is_not_valid'
+  end
+
+  test 'add valid image and display' do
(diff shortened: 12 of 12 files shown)


app/controllers/images_controller.rb
```diff
@@ -0,0 +1,18 @@
+class ImagesController < ApplicationController
+  def new
+    @image = Image.new
+  end
+
+  def create
+    @image = Image.new(params.require(:image).permit(:url))
```
HaominZhang: How about extract `params.require(:image).permit(:url)` into a separate method called `image_params` and assign it to a variable? Then at this line, we can simply use the variable?

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    if @image.save
+      redirect_to image_path(id: @image.id)
+    else
+      render :new, status: :bad_request
```
HaominZhang: I think `:unprocessable_entity` would be a better status code here because the server can actually understand what we are doing but the entity is not well-formed. Here is a description about status code`:unprocessable_entity` 
http://www.restpatterns.org/HTTP_Status_Codes/422_-_Unprocessable_Entity

app/models/image.rb
```diff
@@ -0,0 +1,2 @@
+class Image < ActiveRecord::Base
+end
```
HaominZhang: Since in our database the `url` couldn't be empty, and in migration we set url to be null: false, how about let us add a presence validation here?

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,18 @@
+class ImagesController < ApplicationController
+  def new
+    @image = Image.new
+  end
+
+  def create
+    @image = Image.new(params.require(:image).permit(:url))
+    if @image.save
+      redirect_to image_path(id: @image.id)
```
HaominZhang: Could we some notice to tell the user, the image url was successfully saved?

app/controllers/images_controller.rb
HaominZhang: Similarly to above, how about adding an error message saying that the url is not saved correctly?

app/views/images/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<h1>New Image URL</h1>
+
+<% if @image.errors.any? %>
```
HaominZhang: Good job on showing this error message here!
I think we can make some improvements here as well:
 1. We should try to avoid to include logic in view as much as we can, so how about we just render an error message here and put the logic in our controller method. 
2. From user perspective, they may just want to know if their action succeeded or if there is an error, what is that (in plain words). However, the `errors` method here sometimes gives the error message too technically for users to understand. So how do you think if we failed to save image link, we just give a very simple error message like `Failed to save image url` or if the `url` is empty, then we say `Url can't be blank`?

app/views/images/new.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <ul>
+      <% @image.errors.messages.values.each do |msg| %>
+        <li> <%= msg %> </li>
+      <% end %>
+    </ul>
+<% end %>
+
+<%= form_for :image, url:images_path  do |f| %>
```
HaominZhang: We can add a single space after `url:` for code readability.

app/views/images/show.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>Image Created</h1>
```
HaominZhang: As I commented here https://github.com/AppFolioOnboarding/image-sharer-ragurney/pull/19/files#r75220194, how about we pass a message from controller and then display it in the view, instead of showing this fixed header

HaominZhang:
Right now everything is in one commit (the creation of model, view, and different controller methods). However we should always keep the commit atomic. And for testing, theoretically we should follow the TDD but it is fine that we write tests right after we finish some development. E.g. write unit tests after creating the model, write corresponding controller test after we adding a controller method. So my suggestion is to create several atomic commits, like creating image model, creating `new` method in controller etc, which also contains the corresponding tests. How do you think?

ragurney:
I've gone back and made the changes you suggested, as well as took a stab at changing my commits to make them slightly more atomic.

ragurney:
I just pushed commits which contain my old code as well as the changes you suggested. I also took a stab at editing the commits so as to make them slightly more atomic.

test/models/image_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'test_helper'
+
+class ImageTest < ActiveSupport::TestCase
+  def test_url_valid
```
HaominZhang: Since in this test case we are testing invalid case, how about renaming this to `test_url__invalid`? The double underscore means a subcase

test/models/image_test.rb
```diff
@@ -0,0 +1,8 @@
...
+    img = Image.new
+    refute img.valid?
+  end
```
HaominZhang: How about adding a `test_url__valid` case?

app/controllers/images_controller.rb
```diff
@@ -4,15 +4,23 @@ def new
...
+      render :new, status: :unprocessable_entity
     end
   end
 
   def show
     @image = Image.find(params['id'])
   end
 end
+
+private
```
HaominZhang: This `private` section should be within the class, could you modify it?

ragurney: 7cc068d

test/models/image_test.rb
ragurney: ed93843

test/models/image_test.rb
ragurney: 2b30b8b

HaominZhang:
Good job on addressing my comments! Thank you. And I think there are still something we can improve at this moment. 
1. To me [this commit](https://github.com/AppFolioOnboarding/image-sharer-ragurney/pull/19/commits/b1343b1ae468c1570c16b8ea7df68e5b9be7af6b) is still not atomic enough and the commit message is a little confusing because it contains adding multiple controller methods and tests as well as adding model and migration. Could you please make this commit into several atomic ones?
2. Just for the convenience of your reviewers, at the place where you made changes to address comments, could you please link back to the original comments or under the original comments give a link to the modification you make.

ragurney:
@HaominZhang Just went back and separated the commits as you suggested -- they definitely make more logical sense now! Please let me know if what I did is what you were looking for.

app/views/images/show.html.erb
ragurney: 18ac904397ed2ee8b78814c37ad4e17c68cb48e6

app/views/images/new.html.erb
ragurney: 18ac904397ed2ee8b78814c37ad4e17c68cb48e6

app/controllers/images_controller.rb
ragurney: 18ac904397ed2ee8b78814c37ad4e17c68cb48e6

test/controllers/images_controller_test.rb
```diff
@@ -11,4 +11,9 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_response :ok
     assert_select 'h1', 1, 'New Image URL'
   end
+  test 'controller should create and put correctly' do
```
HaominZhang: Minor: missing an empty line above

HaominZhang:
The commit messages are confusing in this two commits https://github.com/AppFolioOnboarding/image-sharer-ragurney/pull/19/commits/5f89dcedf7dbf63f4bbf7a703689560c316ba628 and https://github.com/AppFolioOnboarding/image-sharer-ragurney/pull/19/commits/464c3248a74471a4271dd74f3c838d6ef901410a, which should be the changes to unit test, not controller test

test/controllers/images_controller_test.rb
```diff
@@ -11,4 +11,9 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    post images_path(image: { url: @url })
+    assert_response :found
+    assert_not_nil Image.find_by(url: @url)
+  end
```
HaominZhang: Could we add another test case here to test the failure case? We can assert whether we get the correct response code

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    assert_response :ok
+    assert_select 'h1', 1, 'New Image URL'
+  end
+
+  test 'should show correct image' do
+    get image_path(@image)
+    assert_response :ok
+    assert_select 'img', 1
+  end
+
```
HaominZhang: How about adding a test case for show incorrect image? We can pass in a invalid id and assert getting corresponding exception

HaominZhang:
@ragurney Thanks for addressing my comments so fast. I left several more specific comments, please take a look. And one more thing is could you please re-order your commits because as what we have right now, adding the model and db is after the creation any controller method however it doesn't make sense. (We should have model/db first and then have following controller actions). Does this make sense?

test/controllers/images_controller_test.rb
ragurney: 0234b7f

test/controllers/images_controller_test.rb
ragurney: 0234b7f

ragurney:
@HaominZhang Just implemented the changes and put the corresponding commit tags.

HaominZhang:
Can we also add flow test for this story?

HaominZhang:
@ragurney Looks good to me. Feel free to merge it!

ragurney:
Great! Thank you again for all of your feedback and help, I appreciate it!

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    @image = Image.new
+  end
+
+  def create
+    @image = Image.new(image_params)
+    if @image.save
+      flash[:success] = 'Url successfully saved!'
+      redirect_to image_path(id: @image.id)
+    else
+      flash[:error] = 'Url cannot be empty'
```
bboe: Does this have any unintended side effects?

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  def show
+    @image = Image.find(params['id'])
+  rescue ActiveRecord::RecordNotFound => _e
```
bboe: Is this part ` => _e` necessary?

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    render plain: 'The image you were looking for was not found!', status: :not_found
```
bboe: I like that you are providing a more-specific error message than the generic 404 page.

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'test_helper'
+
+class ImagesControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @url = 'http://jstsolutions.net/wp-content/themes/realty/lib/images/key_img2.png'
+    @image = Image.create(url: @url)
```
bboe: What if this fails?

test/models/image_test.rb
```diff
@@ -0,0 +1,13 @@
+require 'test_helper'
+
+class ImageTest < ActiveSupport::TestCase
+  def test_url__invalid
+    img = Image.new
```
bboe: What do you think is more easily readable, `img`, or `image`?

test/models/image_test.rb
```diff
@@ -0,0 +1,13 @@
...
+    refute img.valid?
```
bboe: While not immediately obvious, we want to avoid using `refute` and `assert` directly. If you need to use them, I always recommend adding a third argument which is the message to be displayed when the assertion failed. In this particular case you are calling a method that returns a boolean and takes no arguments. The `refute_predicate` and `assert_predicate` functions are perfect for those cases. Please update in your code where appropriate:

    assert_predicate image, :invalid?

Note I usually try to prefer `assert` over `refute` if a method exists that doesn't require a negation operator.

cc @HaominZhang

test/models/image_test.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def test_url__valid
```
bboe: See above

bboe:
@ragurney I added a few comments. Could you please open a new PR to address that feedback. If you see any similar instances throughout your other code, please consider including those updates in that PR as well.

For atomic commits, I think it makes sense to make one commit per type of update.

test/controllers/images_controller_test.rb
ragurney: Switched to using `create!()` so I would get a RecordInvalid error in a later commit.

test/models/image_test.rb
ragurney: I thought `img` would be more readable, as I wanted to differentiate from `Image`. But I think I like 'image` better now I think about it.

test/models/image_test.rb
ragurney: fc0d0c0

test/models/image_test.rb
ragurney: 4a52d3a

app/controllers/images_controller.rb
ragurney: a536962

PR #21, opened 2016-08-24
Add url validation

### LINKED ISSUE #4: Image URL Validation
#### As a user I want to be prevented from submitting invalid image URLs.

__Acceptance criteria__:
- [x] I cannot successfully save an image with an invalid URL.
- [x] An error message is associated with the appropriate input field on
  failure.

__Discussion Topic__:
- [x] What is a valid image URL?

__Dependencies__:
- Save Image Link

ragurney:
Closes #4

### DIFF
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -11,9 +11,8 @@ class ImagesController < ApplicationController
     @image = Image.new(image_params)
     if @image.save
       flash[:success] = 'Url successfully saved!'
-      redirect_to image_path(id: @image.id)
+      redirect_to image_path(@image)
     else
-      flash[:error] = 'Url cannot be empty and must point to an image!'
       render :new, status: :unprocessable_entity
     end
   end
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -13,7 +13,7 @@ class ImagesController < ApplicationController
       flash[:success] = 'Url successfully saved!'
       redirect_to image_path(id: @image.id)
     else
-      flash[:error] = 'Url cannot be empty'
+      flash[:error] = 'Url cannot be empty and must point to an image!'
       render :new, status: :unprocessable_entity
     end
   end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -1,3 +1,15 @@
+require 'uri'
+
 class Image < ActiveRecord::Base
   validates :url, presence: true
+  validate :url_is_valid
+
+  private
+
+  def url_is_valid
+    return unless url.present?
+
+    errors.add(:url, 'must be image') unless
+      (File.extname(url) =~ /^(.png|.gif|.jpg)$/) && (url =~ /^#{URI.regexp}$/)
+  end
 end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -9,7 +9,7 @@ class Image < ActiveRecord::Base
   def url_is_valid
     return unless url.present?
 
-    errors.add(:url, 'must be image') unless
+    errors.add(:url, 'must be a valid url') unless
       (File.extname(url) =~ /^(.png|.gif|.jpg)$/) && (url =~ /^#{URI.regexp}$/)
   end
 end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -1,15 +1,9 @@
 require 'uri'
 
 class Image < ActiveRecord::Base
+  URL_REGEX = %r{ \A(http|https):\/\/[a-z0-9]+([\-\.]{1}[a-z0-9]+)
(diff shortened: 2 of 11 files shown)


app/models/image.rb
```diff
@@ -1,3 +1,15 @@
...
   validates :url, presence: true
+  validate :url_is_valid
+
+  private
+
+  def url_is_valid
+    return unless url.present?
+
+    errors.add(:url, 'must be a valid url') unless
+      (File.extname(url) =~ /^(.png|.gif|.jpg)$/) && (url =~ /^#{URI.regexp}$/)
```
XanderStrike: Check out `validates_format_of` if you haven't already.

Also, `URI.regexp` is more about identifying URL-like things in bodies of text (like comments or blog posts), than it is about determining whether or not that url is _valid_. There are lots of handy regular expressions out there that ensure the validity of a string as a URL that can be modified to make sure that it ends in an image extension.

You can also add the `allow_blank: false` switch so you only need one validation.

ragurney: I just pushed an update with the changes you suggested! 

One thing though, is that I think I may want to leave the separate checks for the null url vs. the invalid url, as it allows me to have more specific error messages. What do you think?

XanderStrike:
👍

bboe:
@ragurney, the commits on this PR are not really atomic.

image

First, the first two commits are one atomic commit. Whenever a feature is added, its tests should be included in the same commit.

Second, style-fixes, while discovered later in time, should be retroactively fixed in the original commit. The time-sensitive exception is if you discovered those issues after code review started. In which case you'll want to add `[TO SQAUSH]` commits to indicate that you'll squash them prior to merging the PR. The same should be true of fixes made as a result of code review feedback.

Third, please avoid using the "merge" flow on your feature branches. Feature branches should be rebased off their parent branch when they need to be updated.

cc @XanderStrike

bboe:
Also, why did you switch to simple form as part of this story? I think it's a great feature, but this story does not depend on it. Doing this refactor is great, but please do that in a separate PR and then rebase your feature branch after that one has been merged.

app/controllers/images_controller.rb
```diff
@@ -11,9 +11,8 @@ def create
     @image = Image.new(image_params)
     if @image.save
       flash[:success] = 'Url successfully saved!'
-      redirect_to image_path(id: @image.id)
+      redirect_to image_path(@image)
```
bboe: Note that `redirect_to @image` should also work.

test/controllers/images_controller_test.rb
```diff
@@ -28,17 +28,19 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'controller should create and put correctly' do
-    post images_path(image: { url: @url })
+    assert_difference 'Image.count' do
+      post images_path(image: { url: @url })
+    end
 
     assert_response :found
```
bboe: Is there a flash message that should be tested for?

test/models/image_test.rb
```diff
@@ -1,13 +1,56 @@
...
+    ['thisis.com/?invalid', 'asdf https://thisis.com?invalid']
   end
 
-  def test_url__valid
-    img = Image.new(url: 'http://placehold.it/120x120&text=image1')
-    assert img.valid?
+  def extensions
+    ['.jpg', '.png', '.gif']
   end
 end
```
bboe: Do these tests cover all the possible success/failure states encompassed by this regex?

    URL_REGEX = %r{ \A(http|https):\/\/[a-z0-9]+([\-\.]{1}[a-z0-9]+)
                  (.*)?(\/)?(.png|.jpg|.gif)\z }x

Going linearly through the list:
* I don't see any test to ensure the URL starts with http or https
* I don't see any tests for `https`.
* Since `h` `t` `t` `p` are fixed patterns, defensive testing would dictate writing a failing test if any one of those characters are missing.
* Each of `://` are required characters. What if one of those is missing?

That's just testing up to `https://`. Please go through the reminder of the URL and do the same. There should be many test cases.

Also what about case?

Please keep this PR separate from the other general fixes PR since it's a little more involved.

cc @XanderStrike

ragurney:
I switched to simple form because of this requirement "An error message is associated with the appropriate input field on failure." I had it working with form helper, but my pair suggested that I just use simple form because it makes it very easy to have the message displayed with the appropriate field and because I would most likely use it later on.

bboe:
> I switched to simple form because of this requirement "An error message is associated with the appropriate input field on failure."

We talked about this in person, but for posterity: Switching to simple form would constitute as a refactor of existing functionality that is necessary for your feature branch. Such refactors should be made in their own pull request, and, once merged, your feature branch can be rebased to include the required functionality.

test/controllers/images_controller_test.rb
ragurney: b6f1eab

PR #22, opened 2016-08-24
Implement seeds for Image

### LINKED ISSUE #5: Image Seeds
#### As a developer I want at least 20 images to be created when I setup the database.

__Acceptance criteria__:
- [ ] After running `rake db:setup` at least 20 images are shown in the
  homepage.
- [ ] When a new app is deployed on heroku (e.g., every review app), that app
  automatically has at least 20 images on its homepage.

__Dependencies__:
- Image Index

ragurney:
Closes #5

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,15 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+url_list = %w(https://static.pexels.com/photos/1518/black-and-white-city-bird-people.jpg
+              http://digital-photography-school.com/wp-content/uploads/2007/02/black-and-white-tips.jpg
+              https://s-media-cache-ak0.pinimg.com/736x/56/5a/9d/565a9d198452ee8884db9240b63b9ee4.jpg
+              http://inspirationfeeed.files.wordpress.com/2014/03/20481.jpg
+              https://www.art-competition.net/images/winners/Black_White/H7_Laurent_Baheux_Lion-portrait-Kenya.jpg)
+
+4.times do
+  url_list.each do |url|
+    Image.create!(url: url)
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,11 +1,3 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
-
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -5,3 +5,15 @@
...
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+url_list = %w(https://static.pexels.com/photos/1518/black-and-white-city-bird-people.jpg
+              http://digital-photography-school.com/wp-content/uploads/2007/02/black-and-white-tips.jpg
+              https://s-media-cache-ak0.pinimg.com/736x/56/5a/9d/565a9d198452ee8884db9240b63b9ee4.jpg
+              http://inspirationfeeed.files.wordpress.com/2014/03/20481.jpg
+              https://www.art-competition.net/images/winners/Black_White/H7_Laurent_Baheux_Lion-portrait-Kenya.jpg)
+
+4.times do
```
mlewisno: Way to know that 4 times 5 is 20! ;) But yeah, clearly this can work - that said I, the user, got bored and lost count scrolling down the list because of the repetition.

A quick note (nothing that needs changing but still good to know) When making a call to `create!` you can pass in an array of hashes so that you only have to make a single call. I don't know if there is any noticeable change in performance, but it makes your intention clearer - which makes it easier for a future reader (me) to know what you're trying to do :)

Oh yeah, and unless you guys are revisiting the movies feature I'd go ahead and remove the commented code above for good measure ;)

Also, way to use %w for your array of strings #winning

bboe: @ragurney While the story didn't specify (I'll make that more clear) I'd really like there to be 20 unique images. It makes testing easier... e.g., I delete an image of an elephant and it's the only one, then I know things are working. The repetition makes QAing a little more difficult.

cc @mlewisno

db/seeds.rb
```diff
@@ -1,7 +1,15 @@
...
-#   Character.create(name: 'Luke', movie: movies.first)
+url_list = %w(https://static.pexels.com/photos/1518/black-and-white-city-bird-people.jpg
+              http://digital-photography-school.com/wp-content/uploads/2007/02/black-and-white-tips.jpg
+              https://s-media-cache-ak0.pinimg.com/736x/56/5a/9d/565a9d198452ee8884db9240b63b9ee4.jpg
+              http://inspirationfeeed.files.wordpress.com/2014/03/20481.jpg
+              https://www.art-competition.net/images/winners/Black_White/H7_Laurent_Baheux_Lion-portrait-Kenya.jpg)
+
+4.times do
+  Image.create!([
+    { url: url_list[0] },
```
bboe: Can you think of a more programatic way to build this list of 5 hashes?

db/seeds.rb
ragurney: 1e7a35e

db/seeds.rb
ragurney: 1e7a35e

PR #25, opened 2016-08-26
Refactor issue 2

ragurney:

### DIFF
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -19,7 +19,7 @@ class ImagesController < ApplicationController
 
   def show
     @image = Image.find(params['id'])
-  rescue ActiveRecord::RecordNotFound => _e
+  rescue ActiveRecord::RecordNotFound
     render plain: 'The image you were looking for was not found!', status: :not_found
   end
 
--- a/db/migrate/20160826181115_change_images_timestamps_not_null.rb
+++ b/db/migrate/20160826181115_change_images_timestamps_not_null.rb
@@ -1,0 +1,6 @@
+class ChangeImagesTimestampsNotNull < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null(:images, :created_at, false, Time.zone.now)
+    change_column_null(:images, :updated_at, false, Time.zone.now)
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,12 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160819175531) do
+ActiveRecord::Schema.define(version: 20160826181115) do
 
   create_table "images", force: :cascade do |t|
     t.string   "url",        null: false
-    t.datetime "created_at"
-    t.datetime "updated_at"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
   end
-
 end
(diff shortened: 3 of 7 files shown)


ragurney:
I forgot to add updates from issue #3. Will change label back after changes are included in this PR.

db/migrate/20160826181115_change_images_timestamps_not_null.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeImagesTimestampsNotNull < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null(:images, :created_at, false, Time.zone.now)
```
bboe: Will this migration work if you already have data in the database?

db/schema.rb
```diff
@@ -10,12 +10,37 @@
...
-ActiveRecord::Schema.define(version: 20160819175531) do
+ActiveRecord::Schema.define(version: 20160826181115) do
 
   create_table "images", force: :cascade do |t|
     t.string   "url",        null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  create_table "taggings", force: :cascade do |t|
```
bboe: Looks like this doesn't quite belong here. Please double check the changed files when opening a pull request to ensure everything looks as it should.

db/migrate/20160826181115_change_images_timestamps_not_null.rb
ragurney: I think so. From what I was reading in the documentation, the optional fourth argument should replace any existing nulls in the database with its value. Is that right?

db/schema.rb
ragurney: I've edited the commit to include only relevant changes. See e9d3443.

db/migrate/20160826181115_change_images_timestamps_not_null.rb
bboe: I actually was not aware of the last argument. Awesome!

app/controllers/images_controller.rb
```diff
@@ -19,7 +19,7 @@ def create
 
   def show
     @image = Image.find(params['id'])
-  rescue ActiveRecord::RecordNotFound => _e
```
bboe: Is there a better name for this commit than "Refactor image controller"? In general "refactor" is pretty vague, and I don't think that necessarily applies to this commit.

test/flows/images_flow_test.rb
```diff
@@ -30,6 +30,6 @@ class ImagesFlowTest < FlowTestCase
```
bboe: We should avoid having commits pertaining to issues with git "Resolve merge conflict". In general, if there are merge conflicts, they should be resolved within whatever atomic commit makes sense. Perhaps this commit was accidentally renamed differently than expected.

bboe:
Added a few additional comments.

test/flows/images_flow_test.rb
ragurney: I believe I accidentally committed the merge resolution rather than just do a git rebase --continue. I've squashed the offending commit into its original (4a32667).

app/controllers/images_controller.rb
ragurney: 6c9764d

test/flows/images_flow_test.rb
bboe: I think this should still be in it's own commit. Since the title of this one is: Rename img to image

bboe:
Almost there.

test/flows/images_flow_test.rb
ragurney: Ok just put it in its own commit: b6ebfa2.

bboe:
👍

PR #26, opened 2016-08-27
Refactored regex to use gem instead

ragurney:


test/models/image_test.rb
```diff
@@ -2,30 +2,8 @@
...
-    invalid_urls.each do |url|
-      extensions.each do |ext|
-        img = Image.new(url: "#{url}#{ext}")
-        refute img.valid?
-        assert_equal ['must be a valid url'], img.errors.messages[:url]
-      end
-    end
-  end
-
-  test 'url invalid with no extension' do
```
bboe: While there is no need to test all invalid cases, there should still be a test that checks for at least one invalid case. This test serves to ensure that the URL validation doesn't accidentally get removed.

bboe:
One test required. Otherwise looks great. 👍

PR #28, opened 2016-09-06
Implement image delete functionality

### LINKED ISSUE #8: Delete Image
#### As a user I want to delete an image.

__Story__:

Our users are awesome at adding images to our web service. Some of our users
add dozens in a single session. Occasionally, however, some users add the wrong
image, and would like a way to remove those images. Please provide our users
with the ability to delete images. Note that at this point we trust our users
not to abuse this feature and delete more images than they should.

__Acceptance criteria__:
- [x] The user is prompted for confirmation prior to deletion.
- [x] Deleted images are removed from the homepage.

__Dependencies__:
- Image Index

ragurney:
Closes #8

### DIFF
--- a/test/flows/images_flow_test.rb
+++ b/test/flows/images_flow_test.rb
@@ -6,7 +6,7 @@ class ImagesFlowTest < FlowTestCase
     @invalid_url = 'asdf https://oh_no.this/?is_not_valid'
   end
 
-  test 'add valid image and display' do
+  test 'add valid image and display, then check deletion functionality' do
     visit(new_image_path)
 
     fill_in('image[url]', with: @valid_url)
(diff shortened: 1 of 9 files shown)


test/flows/images_flow_test.rb
```diff
@@ -19,6 +19,9 @@ class ImagesFlowTest < FlowTestCase
     visit(root_path)
     assert (has_current_path? root_path), 'root path should exist'
     assert has_css?("img[src='#{@valid_url}']"), 'home page should list added image'
+
+    click_link('Delete')
+    assert has_no_css?("img[src='#{@valid_url}']")
```
D-D-M: Maybe you'll want to update the name of this test now that it deletes something. Also, you might want to test out your confirmation modal [here](https://github.com/AppFolioOnboarding/image-sharer-ragurney/pull/28/files#diff-54003fc1b9e144a95e605d0326f6507aR11). Could be useful to:
1. Click the delete link
2. Click "Cancel"
3. Assert that you still have the image there
4. Click delete again
5. Click "Confirm"
6. Assert the image is gone

Your next question might be, "But DDM, I don't currently have to click 'Confirm' when I click the delete button....why isn't the modal popping up?" This might be because you don't have a Javascript driver as your default Capybara driver (Capybara uses `Rack::Test` by default, which can't run Javascript). So you can change that by doing something [like this](https://github.com/AppFolioOnboarding/base/commit/95c16ab53c7064ebfc7fe7e244a974785da4c6c2#diff-e32affa1c612d33896f619dfcede19cfR30). 😃

ragurney: Awesome! I was wondering how to test the modal, thanks! c9b9737

D-D-M:
Nice work, LGTM! 👍

PR #30, opened 2016-09-08
Implement Images by tag

### LINKED ISSUE #9: Images by Tag
#### As a user I want to view all images associated with a tag.

__Story__:

Now that we have added the ability for images to be associated with tags, let
us provide the ability for users to filter images by a given tag. Nana, who
loves fast planes, will be delighted to see only images labeled with SR-71.

__Acceptance criteria__:
- [x] When I click on a tag, I see a filtered list of only the images that have
  the tag I clicked on.
- [x] The tag name, not its ID, is used to perform the filtering.

__Discussion topics__:
- [x] Where in the code should we implement this functionality?
- [x] What happens if a user tries to manually filter by a nonexistent tag?

__Dependencies__:
- Image Tags

ragurney:
Closes #9

### DIFF
--- a/app/business_logic/image_selector.rb
+++ b/app/business_logic/image_selector.rb
@@ -1,0 +1,9 @@
+class ImageSelector
+  def self.select(tag)
+    if tag.nil?
+      Image.order(created_at: :desc)
+    else
+      Image.tagged_with(tag).order(created_at: :desc)
+    end
+  end
+end
--- a/app/views/images/_image_card.html.erb
+++ b/app/views/images/_image_card.html.erb
@@ -2,7 +2,9 @@
   <%= image_tag(image.url, class: 'card-img image-tile img-circle') %>
   <div class = "card-block">
     <% image.tags.each do |tag| %>
-        <span class="tag tag-default image-tag"><%= tag.name %></span>
+        <%= link_to images_path(tag: tag.name) do%>
+          <span class="tag tag-default image-tag"><%= tag.name %></span>
+        <% end %>
     <% end %>
     <div class="card-block">
     <%= link_to 'Delete', image_path(image), method: :delete,
--- a/test/business_logic/image_selector_test.rb
+++ b/test/business_logic/image_selector_test.rb
@@ -1,0 +1,48 @@
+require 'test_helper'
+
+class ImageSelectorTest < ActiveSupport::TestCase
+  test 'should show all images when pass nil for tags' do
+    image3 = Image.create!(url: 'http://www.validurl.com/image2.png', tag_list: nil,
+                           created_at: Time.zone.now)
+    image1 = Image.create!(url: 'http://www.validurl.com/image.png', tag_list: 'tag1, tag2, tag3',
+                           created_at: Time.zone.now - 3.days)
+    image2 = Image.create!(url: 'http://www.validurl.com/image1.png', tag_list: 'tag1, tag2',
+                           created_at: Time.zone.now - 2.days)
+
+    result = ImageSelector.select nil
+    expected = [image3, image2, image1]
+
+    assert_equal expected, result
+  end
+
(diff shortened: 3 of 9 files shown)


app/business_logic/image_selector.rb
```diff
@@ -0,0 +1,9 @@
+class ImageSelector
+  def select(tag)
+    if tag.nil?
+      Image.order(created_at: :desc)
+    else
+      Image.tagged_with(tag).order(created_at: :desc)
```
shiboying: I like the idea of separating this under busines_logic directory. Here, it is better to handle the case that if the tag is not founded in your Image. And also since this class doesn't need any object and what is does is to select the image with the tag pass in. So I would recommend you to change `select` to a class method.

app/views/images/_image_card.html.erb
```diff
@@ -2,7 +2,9 @@
   <%= image_tag(image.url, class: 'card-img image-tile img-circle') %>
   <div class = "card-block">
     <% image.tags.each do |tag| %>
-        <span class="tag tag-default image-tag"><%= tag.name %></span>
+        <%= link_to root_path(tag: tag.name) do%>
```
shiboying: Is that root_path or image_path?

test/business_logic/image_selector_test.rb
```diff
@@ -0,0 +1,33 @@
+require 'test_helper'
+
+class ImageSelectorTest < ActiveSupport::TestCase
+  test 'pass nil for tags' do
```
shiboying: It is not a big deal to me. But I would prefer the format of naming tests as `test 'should show all images when pass nil for tags' do`. In this way, it is easier for people to read and understand what the test is expecting.

test/business_logic/image_selector_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    image2 = Image.create!(url: 'http://www.validurl.com/image1.png', tag_list: 'tag1, tag2',
+                           created_at: Time.zone.now - 2.days)
+
+    result = ImageSelector.new.select nil
+    expected = [image3, image2, image1]
+
+    assert_equal expected, result
+  end
+
+  test 'pass value for tags' do
```
shiboying: You could also add a test for `pass not-exist tag`

shiboying:
Do you have flow test? If so, you should also add tests in flow test.

app/views/images/_image_card.html.erb
ragurney: In this implementation it should be root_path, as when the user clicks on the link the expected behavior is to redirect to the root_path (which is Image index in this case), but passing the tag in the url params so the images can be filtered. I did this so the url would be what I expected: `http://localhost:3000/?tag=lion` instead of `http://localhost:3000/images/?tag=lion`

shiboying: I've looked at your route.rb. It is not the path for images#index is root_path, it is you set root_path to be the the path of images#index. The path for images#index is `http://localhost:3000/images` and since you add the tag filter inside index action, the path should be `http://localhost:3000/images/?tag=lion`. I think you should also add `resources images, only: [:index]` in your route.rb.

ragurney:
Added flow test: 2821a2d

app/views/images/_image_card.html.erb
ragurney: Ah, ok I see. I've made the change: 4e50243

test/business_logic/image_selector_test.rb
ragurney: 3d4a3a5

app/business_logic/image_selector.rb
ragurney: e952002

shiboying:
Look good to me.

PR #32, opened 2016-09-08
Image share no modal

### LINKED ISSUE #10: Share Image
#### As a user I want to share an image with someone via email.

__Story__:

This application is called Image Sharer. Sure, you can share images by adding
them to our web service, but how will people who do not know about this web
service be able to see those images? In phone calls with our customers we have
discovered that many of our users are manually emailing their friends links to
the images. That is awesome! We can make things more efficent for our users by
directly supporting sharing images through email.

__Acceptance criteria__:
- [x] There is a link to share an image.
- [ ] After clicking on this link, a bootstrap modal appears with a form to
  send the email.
- [x] The form has fields for the recipient's email and an optional message.
- [x] The email we send has an HTML part and a text part.
- [x] The email contains the image, the custom message if provided, and a link
  inviting the recipient to check out our application.
- [ ] The modal closes and a success message is shown if the email was sent
  successfully.
- [x] The user is alerted of errors if the operation was unsuccessful.

__Note__: This story is large, so it makes sense to break the implementation
down into more manageable pieces (a la elephant carpaccio). Please start by
implementing a basic sharing feature that uses a new page (not a modal), and
open a PR for that. Then, once we've verified the simplified version is
working, refactor it to use a modal, and open a second PR for that. The
[video](https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax)
provides some background and code examples that will be really helpful for the
second PR, so it's worth watching (and following along) before you begin that
part.

__Discussion topics__:
- [ ] How could we develop the front end and back end separately? (And why
  might that be a good idea?)
- [ ] How should we implement a non-CRUD action?
- [ ] What is AJAX? What libraries can help us add AJAX to our Rails app?
- [ ] How should we go about sending emails from our Rails app?
- [ ] How could we show a preview of the image being shared in the modal?
- [ ] What are the implications of allowing users to send emails from our app?

__Reference material__:
- http://guides.rubyonrails.org/action_mailer_basics.html
- https://elements.heroku.com/addons/sendgrid
- https://devcenter.heroku.com/articles/sendgrid#ruby-rails
- https://github.com/rails/jquery-ujs/wiki
- http://v4-alpha.getbootstrap.com/components/modal
- https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax

__Dependencies__:
- Image Index

ragurney:
First part of #10 (no modal)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ gem 'acts-as-taggable-on', '~> 4.0'
 gem 'bootstrap', '~> 4.0.0.alpha3'
 
 gem 'validate_url'
+gem 'email_validator'
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -1,5 +1,5 @@
 class ImagesController < ApplicationController
-  before_action :set_image, only: [:show, :share_new, :destroy]
+  before_action :set_image, only: [:show, :share_new, :share_send, :destroy]
 
   def index
     @image_list = ImageSelector.select params[:tag]
@@ -25,6 +25,18 @@ class ImagesController < ApplicationController
   def share_new
   end
 
+  def share_send
+    email_info = email_params
+
+    ImageMailer.send_share_email(email_info[:email_address], email_info[:message], @image.url).deliver_now
+  rescue ArgumentError
+    flash[:error] = 'Email cannot be null!'
+    render :share_new, status: :unprocessable_entity
+  else
+    flash[:success] = 'Image successfully sent!'
+    redirect_to root_path
+  end
+
   def destroy
     @image.destroy
     flash[:success] = 'Image successfully deleted!'
@@ -35,6 +47,10 @@ class ImagesController < ApplicationController
 
   def image_params
     params.require(:image).permit(:url, :tag_list)
+  end
+
+  def email_params
+    params.require(:email_info).permit(:email_address, :message)
   end
 
   def set_image
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -28,7 +28,8 @@ class ImagesController < ApplicationController
   def share_send
     email_info = email_params
 
-    ImageMailer.send_share_email(email_info[:email_address], email_info[:message], @image.url).deliver_now
+    ImageMailer.send_share_email(email_address: email_info[:email_address], message: email_info[:message],
+                                 url: @image.url).deliver_now
   rescue ArgumentError
     flash[:error] = 'Email cannot be null!'
     render :share_new, status: :unprocessable_entity
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -1,5 +1,5 @@
 class ImagesController < ApplicationController
-  before_action :set_image, only: [:show, :destroy]
+  before_action :set_image, only: [:show, :share_new, :destroy]
 
   def index
     @image_list = ImageSelector.select params[:tag]
@@ -20,6 +20,9 @@ class ImagesController < ApplicationController
   end
 
   def show
+  end
+
+  def share_new
   end
 
   def destroy
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -59,7 +59,7 @@ class ImagesController < ApplicationController
     if (@image = Image.find_by(id: params[:id]))
       @image
     else
-      flash[:error] = 'The image you were looking for does not exist'
+      flash[:danger] = 'The image you were looking for does not exist'
       redirect_to images_path
     end
   end
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -23,19 +23,20 @@ class ImagesController < ApplicationController
   end
 
   def share_new
+    @share_form = ShareForm.new
   end
 
   def share_send
-    email_info = email_params
+    @share_form = ShareForm.new(email_params)
 
-    ImageMailer.send_share_email(email_address: email_info[:email_address], message: email_info[:message],
-                                 url: @image.url).deliver_now
-  rescue ArgumentError
-    flash[:error] = 'Email cannot be null!'
-    render :share_new, status: :unprocessable_entity
-  else
-    flash[:success] = 'Image successfully sent!'
-    redirect_to root_path
+    if @share_form.valid?
+      ImageMailer.send_share_email(email_address: @share_form.email_address, message: @share_form.message,
+                                   url: @image.url).deliver_now
+      flash[:success] = 'Email successfully sent!'
+      redirect_to root_path
+    else
+      render :share_new, status: :unprocessable_entity
+    end
   end
 
   def destroy
@@ -51,7 +52,7 @@ class ImagesController < ApplicationController
   end
 
   def email_params
-    params.require(:email_info).permit(:email_address, :message)
+    params.require(:share_form).permit(:email_address, :message)
   end
 
   def set_image
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -1,4 +1,6 @@
 class ImagesController < ApplicationController
+  before_action :set_image, only: [:show, :destroy]
+
   def index
     @image_list = ImageSelector.select params[:tag]
   end
@@ -18,14 +20,10 @@ class ImagesController < ApplicationController
   end
 
   def show
-    @image = Image.find(params['id'])
-  rescue ActiveRecord::RecordNotFound
-    render plain: 'The image you were looking for was not found!', status: :not_found
   end
 
   def destroy
-    image = Image.find(params['id'])
-    image.destroy
+    @image.destroy
     flash[:success] = 'Image successfully deleted!'
     redirect_to images_path
   end
@@ -34,5 +32,14 @@ class ImagesController < ApplicationController
 
   def image_params
     params.require(:image).permit(:url, :tag_list)
+  end
+
+  def set_image
+    if (@image = Image.find_by(id: params[:id]))
+      @image
+    else
+      flash[:error] = 'The image you were looking for does not exist'
+      redirect_to images_path
+    end
   end
 end
--- a/app/forms/share_form.rb
+++ b/app/forms/share_form.rb
@@ -1,0 +1,7 @@
+class ShareForm
+  include ActiveModel::Model
+
+  attr_accessor :email_address, :message
+
+  validates :email_address, presence: true, email: true
+end
--- a/app/mailers/image_mailer.rb
+++ b/app/mailers/image_mailer.rb
@@ -1,5 +1,5 @@
 class ImageMailer < ApplicationMailer
-  def send_share_email(email_address, message, url)
+  def send_share_email(email_address:, message:, url:)
     @message = message
     @url = url
 
--- a/app/mailers/image_mailer.rb
+++ b/app/mailers/image_mailer.rb
@@ -1,0 +1,8 @@
+class ImageMailer < ApplicationMailer
+  def send_share_email(email_address, message, url)
+    @message = message
+    @url = url
+
+    mail(to: email_address, from: 'donotreply@appfolio.com', subject: 'Image Shared from Image-Sharer!')
+  end
+end
--- a/app/views/image_mailer/send_share_email.html.erb
+++ b/app/views/image_mailer/send_share_email.html.erb
@@ -1,0 +1,14 @@
+<h1>Hello from Image Sharer!</h1>
+<p>
+  You have been sent an image!
+</p>
+<% if @message.present? %>
+    <p>Your friend said:</p>
+    <p><%= @message %></p>
+<% end %>
+<%= image_tag @url %>
+<p>Thanks for viewing and have a great day!</p>
+
+<p style="opacity: 0.5;"> Want to see more images? Check out our <%= link_to 'website',
+                                                                          root_url %>
+</p>
--- a/app/views/image_mailer/send_share_email.text.erb
+++ b/app/views/image_mailer/send_share_email.text.erb
@@ -1,0 +1,12 @@
+Hello from Image Sharer!
+=========================
+
+You have been sent an image, click <%= @url %> to check it out.
+
+<% if @message.present? %>
+  Your friend said:
+  <%= @message %>
+<% end %>
+Thanks for viewing and have a great day!
+
+Want to see more images? Check out our website: <%= root_url %>
--- a/app/views/images/share_new.html.erb
+++ b/app/views/images/share_new.html.erb
@@ -1,0 +1,18 @@
+<h1 class="text-xs-center">Share Your Image</h1>
+
+<div class="row">
+  <div class="col-xs-6">
+    <%= simple_form_for :email_info, url: share_send_image_path(@image)  do |f| %>
+        <%= f.error_notification  %>
+
+        <%= f.input :email_address, label: "Recipient's email", input_html: { placeholder: 'Enter email'} %>
+        <%= f.input :message, label: 'Message (optional)', as: :text,
+                                                           input_html: { placeholder: 'Enter message'} %>
+        <div class="card text-xs-center">
+          <%= image_tag @image.url, class: 'image-tile card-img' %>
+        </div>
+
+        <%= f.button :submit, 'Send', class: 'btn btn-primary' %>
+    <% end %>
+  </div>
+</div>
--- a/app/views/images/share_new.html.erb
+++ b/app/views/images/share_new.html.erb
@@ -2,7 +2,7 @@
 
 <div class="row">
   <div class="col-xs-6">
-    <%= simple_form_for :email_info, url: share_send_image_path(@image)  do |f| %>
+    <%= simple_form_for @share_form, url: share_send_image_path(@image)  do |f| %>
         <%= f.error_notification  %>
 
         <%= f.input :email_address, label: "Recipient's email", input_html: { placeholder: 'Enter email'} %>
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -24,7 +24,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     get image_path(id: -1)
 
     assert_redirected_to images_path
-    assert_equal 'The image you were looking for does not exist', flash[:error]
+    assert_equal 'The image you were looking for does not exist', flash[:danger]
   end
 
   test 'controller should create and put correctly (just image)' do
@@ -150,7 +150,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
       post share_send_image_path(image, share_form: { email_address: 'valid@valid.com', message: 'Hi!' })
     end
     assert_redirected_to images_path
-    assert_equal 'The image you were looking for does not exist', flash[:error]
+    assert_equal 'The image you were looking for does not exist', flash[:danger]
   end
 
   test 'images are ordered by descending creating time' do
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -23,8 +23,8 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   test 'should show error for incorrect image' do
     get image_path(id: -1)
 
-    assert_response :not_found
-    assert_equal 'The image you were looking for was not found!', response.body
+    assert_redirected_to images_path
+    assert_equal 'The image you were looking for does not exist', flash[:error]
   end
 
   test 'controller should create and put correctly (just image)' do
@@ -124,6 +124,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     end
     assert_redirected_to images_path
     assert_equal 'Image successfully deleted!', flash[:success]
+  end
+
+  test 'delete nonexistent image redirects correctly' do
+    image = Image.create!(url: 'https://static.pexels.com/photos/7919/pexels-photo.jpg')
+    image.destroy
+
+    delete image_path(image)
+    assert_redirected_to images_path
+    assert_equal 'The image you were looking for does not exist', flash[:danger]
   end
 
   private
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -89,6 +89,26 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_select 'h1', 1, 'Images Homepage'
   end
 
+  test 'share new displays correct form' do
+    image = Image.create!(url: 'https://static.pexels.com/photos/7919/pexels-photo.jpg')
+    get share_new_image_path(image)
+
+    assert_response :ok
+    assert_select 'img', 1
+    assert_select 'form[action=?]', share_send_image_path(image)
+    assert_select 'img[src=?]', 'https://static.pexels.com/photos/7919/pexels-photo.jpg'
+    assert_select 'h1', 1, 'Share Your Image'
+  end
+
+  test 'share new missing image displays correct flash message and redirects' do
+    image = Image.create!(url: 'https://static.pexels.com/photos/7919/pexels-photo.jpg')
+    image.destroy
+
+    get share_new_image_path(image)
+    assert_redirected_to images_path
+    assert_equal 'The image you were looking for does not exist', flash[:danger]
+  end
+
   test 'images are ordered by descending creating time' do
     urls = create_images.map(&:url)
 
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -109,6 +109,45 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'The image you were looking for does not exist', flash[:danger]
   end
 
+  test 'share_send action email' do
+    image = Image.create!(url: 'https://example.com')
+    assert_difference 'ActionMailer::Base.deliveries.size', +1 do
+      post share_send_image_path(image), params: { email_info: { email_address: 'example@example.com',
+                                                                 message: 'Hi!' } }
+    end
+    assert_response :found
+    assert_equal 'Image successfully sent!', flash[:success]
+
+    share_email = ActionMailer::Base.deliveries.last
+    assert_equal 'Image Shared from Image-Sharer!', share_email.subject
+    assert_equal 'example@example.com', share_email.to[0]
+
+    ['Hello from Image Sharer!', image.url, 'Hi!'].each do |value|
+      assert_match value, share_email.html_part.to_s
+      assert_match value, share_email.text_part.to_s
+    end
+  end
+
+  test 'share_send action no email error' do
+    image = Image.create!(url: 'https://example.com')
+    assert_difference 'ActionMailer::Base.deliveries.size', 0 do
+      post share_send_image_path(image), params: { email_info: { email_address: '',
+                                                                 message: '' } }
+    end
+    assert_response :unprocessable_entity
+    assert_equal 'Email cannot be null!', flash[:error]
+  end
+
+  test 'share_send action image does not exist' do
+    image = Image.create!(url: 'https://example.com')
+    image.destroy
+    assert_difference 'ActionMailer::Base.deliveries.size', 0 do
+      post share_send_image_path(image, share_form: { email_address: 'valid@valid.com', message: 'Hi!' })
+    end
+    assert_redirected_to images_path
+    assert_equal 'The image you were looking for does not exist', flash[:error]
+  end
+
   test 'images are ordered by descending creating time' do
     urls = create_images.map(&:url)
 
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -180,6 +180,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  test 'check images have share button and correct link' do
+    image = Image.create!(url: 'http://validurl.com', tag_list: 'some, awesome, tags')
+
+    get image_path(image)
+
+    assert_response :ok
+    assert_select '.js-share-image[href=?]', "/images/#{image.id}/share_new"
+  end
+
   test 'delete removes image successfully' do
     image = Image.create!(url: 'https://www.google.com/image2.png')
 
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -112,11 +112,10 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   test 'share_send action email' do
     image = Image.create!(url: 'https://example.com')
     assert_difference 'ActionMailer::Base.deliveries.size', +1 do
-      post share_send_image_path(image), params: { email_info: { email_address: 'example@example.com',
-                                                                 message: 'Hi!' } }
+      post share_send_image_path(image, share_form: { email_address: 'example@example.com', message: 'Hi!' })
     end
     assert_response :found
-    assert_equal 'Image successfully sent!', flash[:success]
+    assert_equal 'Email successfully sent!', flash[:success]
 
     share_email = ActionMailer::Base.deliveries.last
     assert_equal 'Image Shared from Image-Sharer!', share_email.subject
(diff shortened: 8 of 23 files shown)


rmacklin:
In this case you don't want to say "Closes #10" because this is only the first part, so you don't want the issue to be automatically closed when this PR is merged. You can edit the description to say "First part of #10 (no modal)" and it will still link to the issue without automatically closing it upon merge.

rmacklin:
Also, I'm bringing in @HaominZhang as a second reviewer for this story so he will make some comments too.

app/mailers/image_mailer.rb
```diff
@@ -0,0 +1,8 @@
+class ImageMailer < ApplicationMailer
+  def send_share_email(email_address, message, url)
```
HaominZhang: Can we add `default from` email here, using `donotreply@appfolio.com`

app/views/image_mailer/send_share_email.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1>Hello from Image Sharer!</h1>
+<p>
+  You have been sent an image!
+</p>
+<p> <%= @message %> </p>
```
HaominZhang: If the `message` is optional, do we still want to display it? Is there a better way to handle the empty message case?

app/views/image_mailer/send_share_email.text.erb
```diff
@@ -0,0 +1,10 @@
+Hello from Image Sharer!
+=========================
+
+You have been sent an image, click <%= @url %> to check it out.
+
+<%= @message %>
```
HaominZhang: Same as above

test/mailers/previews/image_mailer_preview.rb
```diff
@@ -0,0 +1,3 @@
+# Preview all emails at http://localhost:3000/rails/mailers/image_mailer
+class ImageMailerPreview < ActionMailer::Preview
```
HaominZhang: Can we have a method that create a sample email and call `send_share_email ` to send the sample email out?

app/controllers/images_controller.rb
```diff
@@ -18,9 +20,11 @@ def create
   end
 
   def show
-    @image = Image.find(params['id'])
-  rescue ActiveRecord::RecordNotFound
-    render plain: 'The image you were looking for was not found!', status: :not_found
+  end
```
HaominZhang: Can we make the changing of this part in a separate commit that is about using before action to find image?

app/controllers/images_controller.rb
```diff
@@ -18,9 +20,11 @@ def create
...
+
+  def share_new
+  end
+
   end
 
   def destroy
```
HaominZhang: Since we are using before_action, do we still want to find image here? Can we add `destroy` method to before_action as well?

test/controllers/images_controller_test.rb
```diff
@@ -79,6 +79,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_select 'h1', 1, 'Images Homepage'
   end
 
+  test 'share new displays correct form' do
+    image = Image.create!(url: 'https://static.pexels.com/photos/7919/pexels-photo.jpg')
+    get share_new_image_path(image)
+
+    assert_response :ok
+    assert_select 'img', 1
+    assert_select 'h1', 1, 'Share Your Image'
```
HaominZhang: Can we also assert the image is exactly what we created?

test/controllers/images_controller_test.rb
```diff
@@ -79,6 +79,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  end
+
```
HaominZhang: Can we all have a test case for the behavior if image doesn’t exist?

app/controllers/images_controller.rb
```diff
@@ -25,6 +25,17 @@ def show
   def share_new
   end
 
+  def share_send
+    image_url = @image.url
+    email_info = email_params
+
+    ImageMailer.send_share_email(email_info[:email_address], email_info[:message], image_url).deliver_now
+  rescue ArgumentError
+    flash[:error] = 'Email cannot be null!'
```
HaominZhang: How about the email address is invalid format?

HaominZhang:
The share_email itself is more like an model object that we can have a separate class for it and then for its attributes(email, message) we can have different validations on them, e.g. for email, we can have both presence and format validation. Can you add a class for it?

app/mailers/image_mailer.rb
rmacklin: How about using keyword arguments for this to make the invocations self-documenting (and so that parameter order doesn't matter in the invocations)?

app/views/image_mailer/send_share_email.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<%= image_tag @url %>
+<p>Thanks for viewing and have a great day!</p>
+
+<p class="text-muted"> Want to see more images? Check out our <%= link_to 'website',
```
rmacklin: This class isn't defined in your emails so adding it will have no effect. For a simple style, you can just declare it inline with the `style` attribute.

app/views/image_mailer/send_share_email.html.erb
```diff
@@ -0,0 +1,11 @@
...
+                                                                          'https://ragurney.herokuapp.com/' %>
```
rmacklin: We should avoid hardcoding this URL in the code. If your application's site changed, we wouldn't want to have to redeploy the application to make our emails have the right URL. Instead, we can use `root_url` here and set the `host` with an environment variable in the production environment.

app/views/image_mailer/send_share_email.text.erb
```diff
@@ -0,0 +1,10 @@
...
+
+Thanks for viewing and have a great day!
+
+Want to see more images? Check out our website: https://ragurney.herokuapp.com/
```
rmacklin: Same as above

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,16 @@
...
+    email = ImageMailer.send_share_email('me@example.com', 'Hello!', 'https://isexample.com/image.png')
+
+    assert_emails 1 do
+      email.deliver_now
+    end
+
+    assert_equal ['me@example.com'], email.to
+    assert_equal 'Image Shared from Image-Sharer!', email.subject
+    assert_equal read_fixture('send_share_email.text').join, email.body.parts[0].body.to_s
+    assert_equal read_fixture('send_share_email.html').join, email.body.parts[1].body.to_s
```
rmacklin: This works, but we tend to avoid using email fixtures. They add an unnecessary level of indirection to the test that makes it harder to see what is being tested, and you can still make the same assertions without using fixtures.

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+end
```
rmacklin: What about a test for the case when a message wasn't provided?

app/controllers/images_controller.rb
```diff
@@ -34,4 +38,10 @@ def destroy
   def image_params
     params.require(:image).permit(:url, :tag_list)
   end
+
+  def set_image
+    @image = Image.find(params['id'])
+  rescue ActiveRecord::RecordNotFound
+    render plain: 'The image you were looking for was not found!', status: :not_found
```
rmacklin: Rendering this plain text is not the best UX. The user will jump out of our app to a white screen with a small message. A better experience would be to redirect them to the images index with a flash message.

app/controllers/images_controller.rb
```diff
@@ -34,4 +38,10 @@ def destroy
   def image_params
     params.require(:image).permit(:url, :tag_list)
   end
+
+  def set_image
+    @image = Image.find(params['id'])
```
rmacklin: Convention is to use symbols for accessing params (`:id` rather than `'id'`)

app/controllers/images_controller.rb
rmacklin: Additionally, `rescue` should be reserved for _exceptional_ behavior rather than for normal control flow. We expect that users will run into this code path, so we should handle it with regular conditional logic (`if/else`) rather than exception handling (`rescue`). You can use `Image.find_by(id: params[:id])` instead which will return `nil` rather than raising an exception when the image is not found.

app/views/images/share_new.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1 class="text-xs-center">Share Your Image</h1>
+
+<div class="row">
+  <div class="col-xs-6">
+    <%= simple_form_for :email_info, url: share_send_image_path(@image)  do |f| %>
+        <%= f.error_notification  %>
```
rmacklin: How is this used?

test/controllers/images_controller_test.rb
rmacklin: We should also assert that the `form` is displayed (as the test name suggests)

app/controllers/images_controller.rb
```diff
@@ -25,6 +25,17 @@ def show
   def share_new
   end
 
+  def share_send
+    image_url = @image.url
```
rmacklin: I don't think this variable is necessary.

app/controllers/images_controller.rb
```diff
@@ -25,6 +25,17 @@ def show
...
+    email_info = email_params
+
+    ImageMailer.send_share_email(email_info[:email_address], email_info[:message], image_url).deliver_now
+  rescue ArgumentError
+    flash[:error] = 'Email cannot be null!'
+    render :share_new, status: :unprocessable_entity
+  else
+    flash[:success] = 'Image successfully sent!'
+    redirect_to root_path
```
rmacklin: `rescue/else` is an antipattern that should be avoided. Put the "go-right" case before the `rescue` instead. Also, rescuing `ArgumentError` is too broad and should be avoided.

As Haomin was mentioning in https://github.com/AppFolioOnboarding/image-sharer-ragurney/pull/32#issuecomment-245757839, a better approach would be to create a form object for the share form and have that form object use `ActiveModel` to enforce validations for this form (e.g. require a valid email address).

test/controllers/images_controller_test.rb
```diff
@@ -88,6 +88,32 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'share_send action no email error' do
+    image = Image.create!(url: 'https://example.com')
+    assert_difference 'ActionMailer::Base.deliveries.size', 0 do
+      post share_send_image_path(image), params: { email_info: { email_address: '',
+                                                                 message: '' } }
+    end
+    assert_response :unprocessable_entity
+    assert_equal 'Email cannot be null!', flash[:error]
+  end
+
```
rmacklin: We should also test the case when the image being shared does not exist.

test/controllers/images_controller_test.rb
```diff
@@ -88,6 +88,32 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+                                                                 message: 'Hi!' } }
+    end
+    assert_response :found
+    assert_equal 'Image successfully sent!', flash[:success]
+
+    share_email = ActionMailer::Base.deliveries.last
+    assert_equal 'Image Shared from Image-Sharer!', share_email.subject
+    assert_equal 'example@example.com', share_email.to[0]
+    assert_match 'Hello from Image Sharer!', share_email.body.parts[0].to_s
+    assert_match 'Hello from Image Sharer!', share_email.body.parts[1].to_s
```
rmacklin: We should assert that the correct image was shared. We can also assert that the message was included.

test/mailers/image_mailer_test.rb
rmacklin: Also, it's better to call `.text_part` and `.html_part` explicitly than to get them implicitly via `parts[0]` and `parts[1]`

test/flows/images_flow_test.rb
```diff
@@ -32,6 +32,29 @@ class ImagesFlowTest < FlowTestCase
     assert has_no_css?("img[src='#{@valid_url}']")
   end
 
+  test 'add image then share it' do
+    visit new_image_path
+
+    fill_in('image[url]', with: @valid_url)
+    click_button('Upload Image')
```
rmacklin: It's not necessary to upload an image through the UI in this test. Only the tests that are explicitly exercising the new image form should do that. Every other test should just set up images in the database directly, which will be much faster.

rmacklin: Also, we should create a second image here with a different URL so that we can ensure the share page displays the correct image.

rmacklin:
Did you test your review app? Sharing images doesn't work.

rmacklin:
I noticed you don't have a mailer layout. It's a good idea to add one to ensure all your emails will have a consistent layout wrapper.

rmacklin:
To make sharing images work in production, you'll need to configure Sendgrid, as linked in the issue references.

We can actually deliver the ability to send emails in its own PR and then rebase this one on top of that. The benefits of doing this are:

1. By delivering the ability to send emails separately, you can fully focus on and test the email setup in your staging and production environments without it being tied to the "share image" feature; moreover, other features that require email deliver can then be built without waiting for the "share image" branch to be merged.
2. Each PR is smaller in scope and more focused and thus, more easily testable, more easily reviewable, and less likely to introduce bugs. This all leads to a shorter turnaround time for each PR. Furthermore, if the sharing feature had a bug, you could revert the "share image" PR without reverting the ability to send emails.

rmacklin:
Regarding
> you can fully test the email setup in your staging and production environments without it being tied to the "share image" feature

You may be wondering how this can be tested without a feature that actually sends emails. The answer is: your good friend, the rails console. You can run the console on your review app to create a test mailer and send yourself an email from the review app. Once that's confirmed to work, you can merge the PR and do the same on your staging application, and after verifying it there you can promote staging to production and finally test it there.

rmacklin:
Lastly, if you're wondering why we didn't create a separate story for the "ability to send emails" deliverable, the reason is that this mimics real development. It's a natural part of the process to create a user story, start implementing it, and then realize that you can split out one or more separate deliverables and merge some changes sooner than others. If we had made these separate stories from the beginning, you wouldn't get to develop this skill yourself.

app/mailers/image_mailer.rb
ragurney: 94333c9, and 84c2713 respectively

app/views/image_mailer/send_share_email.html.erb
ragurney: 5223c52

app/views/image_mailer/send_share_email.text.erb
ragurney: 5223c52

test/mailers/previews/image_mailer_preview.rb
ragurney: 0a5bf01

app/controllers/images_controller.rb
ragurney: f774cc0

test/controllers/images_controller_test.rb
ragurney: 37d5f36

test/controllers/images_controller_test.rb
ragurney: cea65b0

app/views/image_mailer/send_share_email.html.erb
ragurney: 4bb7222

app/views/image_mailer/send_share_email.html.erb
ragurney: af21162

app/views/image_mailer/send_share_email.text.erb
ragurney: af21162

test/mailers/image_mailer_test.rb
ragurney: ad75dac

test/mailers/image_mailer_test.rb
ragurney: ad75dac

app/controllers/images_controller.rb
ragurney: cea65b0

app/controllers/images_controller.rb
ragurney: 7fd7581

app/views/images/share_new.html.erb
ragurney: This is used by SimpleForm to display the helpful 'Please review the problems below:' [flash message](https://puu.sh/rdDiP/ecacbea7e7.png)

test/flows/images_flow_test.rb
ragurney: 53a9d55

app/controllers/images_controller.rb
ragurney: 53a9d55

app/controllers/images_controller.rb
ragurney: bf26d9a

test/controllers/images_controller_test.rb
ragurney: 995f3a3

test/controllers/images_controller_test.rb
ragurney: ecb9396

test/flows/images_flow_test.rb
ragurney: This will be addressed in a separate PR which will have the `flowTestWithPageObjects` refactor included. I will link the fix once complete.

app/views/image_mailer/send_share_email.html.erb
```diff
@@ -2,7 +2,10 @@
 <p>
   You have been sent an image!
 </p>
-<p> <%= @message %> </p>
+<% unless @message.empty? %>
```
rmacklin: This works, but when possible, it's preferable to use `if` rather than `unless` (to avoid "double negative"-y conditions). In this case we could use `if @message.present?`

test/mailers/image_mailer_test.rb
```diff
@@ -29,7 +31,56 @@ class ImageMailerTest < ActionMailer::TestCase
...
+    assert_equal expected_email_body_text('', 'https://isexample.com/image.png'),
+                 email.text_part.body.to_s
+  end
+
+  private
+
+  def expected_email_body_html(message, url)
+    message_block = message.empty? ? '' : "    <p>Your friend said:</p>\n    <p>#{message}</p>\n"
+
+    <<~HEREDOC
```
rmacklin: You can actually use anything for the heredoc demarcations. It's conventional to use `HTML` when the string is HTML, `SQL` when it's SQL, `JS` when it's javascript, etc.

test/mailers/image_mailer_test.rb
```diff
@@ -29,7 +31,56 @@ class ImageMailerTest < ActionMailer::TestCase
...
+
+      </body>
+    </html>
+    HEREDOC
+  end
+
+  def expected_email_body_text(message, url)
+    message_block = message.empty? ? '' : "  Your friend said:\n  #{message}\n"
+
+    <<~HEREDOC
```
rmacklin: Per the above, you can use `TEXT` here

Gemfile
```diff
@@ -29,6 +29,7 @@ gem 'acts-as-taggable-on', '~> 4.0'
 gem 'bootstrap', '~> 4.0.0.alpha3'
 
 gem 'validate_url'
+gem 'email_validator'
```
rmacklin: Adding this gem can be its own atomic commit separate from using it in the `ShareForm`.

app/mailers/image_mailer.rb
```diff
@@ -3,6 +3,6 @@ def send_share_email(email_address, message, url)
     @message = message
     @url = url
 
-    mail(to: email_address, subject: 'Image Shared from Image-Sharer!')
+    mail(to: email_address, from: 'donotreply@appfolio.com', subject: 'Image Shared from Image-Sharer!')
```
rmacklin: You probably want this to apply to all emails by default, so you can specify it in the base `ApplicationMailer` class:
```
class ApplicationMailer < ActionMailer::Base
  default from: 'Image-Tiles <donotreply@appfolio.com>'

  layout 'mailer'
end
```

app/views/images/share_new.html.erb
rmacklin: Got it. Thanks

ragurney:
Thank you for the feedback! Some things: 

> And I noticed that the share form appears to go right against the edge of the page: 

This is addressed in my `add-style-extra-exercises` branch, which is currently still in the process of being reviewed. The form will have similar spacing to [this one](https://ragurneys-pr-29.herokuapp.com/images/new) once the PR is merged in. 

> I noticed you don't have a mailer layout. It's a good idea to add one to ensure all your emails will have a consistent layout wrapper.

Are these what you are talking about: [html](https://github.com/AppFolioOnboarding/image-sharer-ragurney/blob/master/app/views/layouts/mailer.html.erb) [text](https://github.com/AppFolioOnboarding/image-sharer-ragurney/blob/master/app/views/layouts/mailer.text.erb)? I think these were included in the base project which is why they are not listed as changes in this PR.

app/mailers/image_mailer.rb
ragurney: a30aa9f

test/mailers/image_mailer_test.rb
ragurney: c40bae1

Gemfile
ragurney: cd10fbc

test/mailers/image_mailer_test.rb
ragurney: c40bae1

app/views/image_mailer/send_share_email.html.erb
ragurney: 06f2fd3

rmacklin:
> Are these what you are talking about: html text? I think these were included in the base project which is why they are not listed as changes in this PR.

Yes. Thanks for pointing that out. I think that must be new in the Rails 5 generator because previously new hires had to add this manually.

rmacklin:
> This is addressed in my add-style-extra-exercises branch, which is currently still in the process of being reviewed.

I see. Perhaps that fix could be split out of that branch into a separate PR (I don't think it's one of the extra exercises).

ragurney:
The addition of page margins has been cherry-picked out of `add-style-extra-exercises` and put into its own PR: #40

test/flows/images_flow_test.rb
ragurney: Fix is now in #41

test/flows/images_crud_test.rb
```diff
@@ -86,4 +86,37 @@ class ImagesCrudTest < FlowTestCase
     images_index_page = images_index_page.clear_tag_filter!
     assert_equal 3, images_index_page.images.count
   end
+
+  test 'share image' do
```
rmacklin: Sharing is not part of the CRUD operations, so we shouldn't add this test in this file. We can create a separate flow test file which will hold all our sharing flow tests.

test/flows/images_crud_test.rb
```diff
@@ -86,4 +86,37 @@ class ImagesCrudTest < FlowTestCase
...
+      { url: happening_url, tag_list: 'its, really, happening' },
+      { url: mind_blown_url, tag_list: 'mind, totally, blown' }
+    ])
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+
+    image_to_share = images_index_page.images.find do |image|
+      image.url == mind_blown_url
+    end
+    image_show_page = image_to_share.view!
```
rmacklin: Do we need this step? Can we just say `share_new_image_page = image_to_share.open_share_page!`?

test/page_objects/images/image_card.rb
```diff
@@ -23,6 +23,11 @@ def delete_and_confirm!
         delete(&:accept)
         window.change_to(IndexPage)
       end
+
+      def click_share!
```
rmacklin: A better name for this is `open_share_page!` (since it does more than just click the share button)

test/flows/images_crud_test.rb
rmacklin: Additionally, can you think of another sharing test that we should have?

ragurney: 7daa5ec, and possibly cf36b40?

test/page_objects/images/image_card.rb
ragurney: bfa57d6

test/flows/images_crud_test.rb
ragurney: f5b8e53

test/flows/share_image_test.rb
```diff
@@ -14,9 +14,8 @@ class ShareImageTest < FlowTestCase
     image_to_share = images_index_page.images.find do |image|
       image.url == mind_blown_url
     end
-    image_show_page = image_to_share.view!
 
-    share_new_image_page = image_show_page.open_share_page!
+    share_new_image_page = image_to_share.open_share_page!
```
rmacklin: 👍  Now we're testing sharing from the index page. I don't think it's necessary to add a flow test for sharing from the show page, but we could test in our controller test that the share button is rendered on the show page and has the correct link. This is similar to what you did in 88a855d713c5d135c23ee43adab332ee0216e9c8

test/flows/share_image_test.rb
```diff
@@ -31,5 +31,15 @@ class ShareImageTest < FlowTestCase
     share_new_image_page.message.set('Some message')
     images_index_page = share_new_image_page.share_image!
     assert_equal 'Email successfully sent!', images_index_page.flash_message(:success)
+
+    image_to_share = images_index_page.images.find do |image|
```
rmacklin: Let's make this its own test case.

ragurney: 6143b56

test/flows/share_image_test.rb
ragurney: 14193db

rmacklin: That checks the share buttons on the index page. What I was saying in this comment was that since we've made this flow test exercise sharing from the index page, we could test that the share button is rendered properly in the controller test for the show action.

ragurney: Oh, of course. Sorry I should have read more carefully. Fix incoming.

ragurney: 0945a22

ragurney:
Changes ready for review.

test/mailers/previews/image_mailer_preview.rb
rmacklin: Does this work on your application? I remember Peter had to modify some CSP stuff in development to get the maiiler preview to work. You may want to check with him if it's not working.

test/flows/share_image_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  test 'share nonexistent image' do
+    happening_url = 'https://media.giphy.com/media/rl0FOxdz7CcxO/giphy.gif'
+    created_image = Image.create!(url: happening_url, tag_list: 'its, really, happening')
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+    image_to_share = images_index_page.images.find do |image|
+      image.url == happening_url
+    end
+    share_new_image_page = image_to_share.open_share_page!
+    created_image.destroy
```
rmacklin: It's better to use the exception-raising `destroy!` so we'll know immediately if the operation failed.

test/flows/share_image_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    share_new_image_page.email_address.set('valid@email.com')
+    share_new_image_page.message.set('Some message')
+    images_index_page = share_new_image_page.share_image!
+    assert_equal 'The image you were looking for does not exist', images_index_page.flash_message(:danger)
+  end
```
rmacklin: Using newlines can improve flow test readability. Consider:
```ruby
happening_url = 'https://media.giphy.com/media/rl0FOxdz7CcxO/giphy.gif'
created_image = Image.create!(url: happening_url, tag_list: 'its, really, happening')

images_index_page = PageObjects::Images::IndexPage.visit
image_to_share = images_index_page.images.find do |image|
  image.url == happening_url
end

share_new_image_page = image_to_share.open_share_page!
created_image.destroy!

share_new_image_page.email_address.set('valid@email.com')
share_new_image_page.message.set('Some message')

images_index_page = share_new_image_page.share_image!
assert_equal 'The image you were looking for does not exist', images_index_page.flash_message(:danger)
```

app/forms/share_form.rb
```diff
@@ -0,0 +1,7 @@
+class ShareForm
+  include ActiveModel::Model
+
+  attr_accessor :email_address, :message
+
+  validates :email_address, email: true
```
rmacklin: We should an explicit presence validation so that a blank submission will result in the error message "can't be blank" rather than "is invalid", and so that simple_form will automatically know the field is required (and style it as such).

test/mailers/previews/image_mailer_preview.rb
ragurney: 3b36f80

test/flows/share_image_test.rb
ragurney: e521f3e

test/flows/share_image_test.rb
ragurney: 34a8d44

app/forms/share_form.rb
ragurney: 19fecee

app/forms/share_form.rb
```diff
@@ -3,5 +3,6 @@ class ShareForm
 
   attr_accessor :email_address, :message
 
+  validates :email_address, presence: true
```
rmacklin: You can combine these:
```ruby
validates :email_address, presence: true, email: true
```

app/controllers/images_controller.rb
```diff
@@ -34,4 +39,13 @@ def destroy
   def image_params
     params.require(:image).permit(:url, :tag_list)
   end
+
+  def set_image
```
rmacklin: This method is currently added in a743e130585c791f66ab3655c0f3c15647e79cd0 but first referenced in an earlier commit, 0a64cbc2b84cabeb8930331a2918da65248953eb. That should be fixed so that the commits prior to this one aren't broken.

app/forms/share_form.rb
ragurney: e8ec90f

ragurney:
The most recent commits have now been squashed and the history has been fixed.

app/controllers/images_controller.rb
```diff
@@ -22,6 +22,11 @@ def create
   def show
   end
 
+  def share_new
+  end
+
+  end
```
rmacklin: This commit (c59178a3d1fe223b18da1cd5fb0b51e86215f846) is broken as there's a stray `end` here.

app/controllers/images_controller.rb
```diff
@@ -1,4 +1,6 @@
 class ImagesController < ApplicationController
+  before_action :set_image, only: [:show, :share_new, :share_send, :destroy]
```
rmacklin: The `:share_new, :share_send` actions don't exist at this point in the history (d0c06233b7506143b6752527dbc7049eecb105b5)

test/controllers/images_controller_test.rb
```diff
@@ -23,8 +23,8 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   test 'should show error for incorrect image' do
     get image_path(id: -1)
 
-    assert_response :not_found
-    assert_equal 'The image you were looking for was not found!', response.body
+    assert_redirected_to images_path
+    assert_equal 'The image you were looking for does not exist', flash[:error]
```
rmacklin: This change is part of c59178a3d1fe223b18da1cd5fb0b51e86215f846 but should be in d0c06233b7506143b6752527dbc7049eecb105b5. Additionally, I'd expect to see a test change for `destroy` as well, since that was added to the `before_action`.

test/controllers/images_controller_test.rb
```diff
@@ -100,14 +100,22 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
   end
 
+  test 'share new missing image displays correct flash message and redirects' do
+    image = Image.create!(url: 'https://static.pexels.com/photos/7919/pexels-photo.jpg')
+    image.destroy
+
+    get share_new_image_path(image)
+    assert_redirected_to images_path
+    assert_equal 'The image you were looking for does not exist', flash[:error]
+  end
```
rmacklin: This test is added in 7903c54e57ec44b810e32a0788f7df8b7a6b9eb0 but should be in d0c06233b7506143b6752527dbc7049eecb105b5

test/flows/share_image_test.rb
```diff
@@ -0,0 +1,54 @@
+require 'flow_test_helper'
+
+class ShareImageTest < FlowTestCase
```
rmacklin: This test is added in 7903c54e57ec44b810e32a0788f7df8b7a6b9eb0, but should be part of the "Write flow test" commit (1c7432676cd6357951c048515ae881512800b543)

app/controllers/images_controller.rb
ragurney: e407752

app/controllers/images_controller.rb
ragurney: I now have this commit 106851c without `:share_new` or `:share_send`, then add them in the subsequent respective commits.

test/controllers/images_controller_test.rb
ragurney: 106851c

test/controllers/images_controller_test.rb
ragurney: This test will be added to 60a9c89 shortly to fix commit history.

test/flows/share_image_test.rb
ragurney: d71f362

test/controllers/images_controller_test.rb
ragurney: Added to 40129a5 (new share_new commit)

PR #35, opened 2016-09-09
Add classes to html tags for easy selection

ragurney:


kkajla12:
👍

app/views/images/_image_card.html.erb
```diff
@@ -1,11 +1,15 @@
...
+  <%= link_to image_tag(image.url, class: 'card-img image-tile img-circle'),
+              image_path(id: image.id),
+              class: 'link_to_image'%>
   <div class = "card-block">
-    <% image.tags.each do |tag| %>
-        <%= link_to images_path(tag: tag.name) do%>
-          <span class="tag tag-default image-tag"><%= tag.name %></span>
-        <% end %>
-    <% end %>
+    <div class="js-image-tags">
```
rmacklin: This div is unnecessary as the `js-image-tags` class can be added on the parent div

app/views/images/_image_card.html.erb
```diff
@@ -1,11 +1,15 @@
-<div class="text-xs-center card" data-image-id="<%= image.id %>">
-  <%= image_tag(image.url, class: 'card-img image-tile img-circle') %>
+<div class="js-image-card-container text-xs-center card" data-image-id="<%= image.id %>">
+  <%= link_to image_tag(image.url, class: 'card-img image-tile img-circle'),
+              image_path(id: image.id),
+              class: 'link_to_image'%>
```
rmacklin: This class name should be something like `image-link`. Classes shouldn't use underscores as the word separator

app/views/images/_image_card.html.erb
```diff
@@ -1,11 +1,15 @@
...
   <div class = "card-block">
```
rmacklin: Avoid putting spaces around the `=` between html attribute names and values.

app/views/images/_image_card.html.erb
```diff
@@ -1,11 +1,15 @@
-<div class="text-xs-center card" data-image-id="<%= image.id %>">
-  <%= image_tag(image.url, class: 'card-img image-tile img-circle') %>
+<div class="js-image-card-container text-xs-center card" data-image-id="<%= image.id %>">
+  <%= link_to image_tag(image.url, class: 'card-img image-tile img-circle'),
```
rmacklin: Adding this link should have been its own atomic commit

PR #36, opened 2016-09-10
Move unecessary div, rename image_link class, and fix '=' spacing

ragurney:
Fixed issues with previous refactor to _image_card fragment.

### DIFF
--- a/app/views/images/_image_card.html.erb
+++ b/app/views/images/_image_card.html.erb
@@ -1,15 +1,13 @@
 <div class="js-image-card-container text-xs-center card" data-image-id="<%= image.id %>">
   <%= link_to image_tag(image.url, class: 'card-img image-tile img-circle'),
               image_path(id: image.id),
-              class: 'link_to_image'%>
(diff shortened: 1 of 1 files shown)


app/views/images/_image_card.html.erb
```diff
@@ -1,15 +1,13 @@
...
-              class: 'link_to_image'%>
-  <div class = "card-block">
-    <div class="js-image-tags">
-      <% image.tags.each do |tag| %>
-          <%= link_to images_path(tag: tag.name) do%>
-            <span class="tag tag-default image-tag"><%= tag.name %></span>
-          <% end %>
-      <% end %>
-    </div>
+              class: 'image-link'%>
```
rmacklin: Actually if this class is only used for selection in tests, you should give is `js-` prefix to make it clear that it's not a style class.

ragurney: 1cfbf44

rmacklin:
Looks good. You can squash and then merge

PR #37, opened 2016-09-13
Refactor delete to add redirect from show page

ragurney:


test/flows/images_flow_test.rb
```diff
@@ -29,6 +29,17 @@ class ImagesFlowTest < FlowTestCase
...
     dialog.accept
+    assert page.has_current_path? images_path
+    assert page.has_content?('Image successfully deleted!')
+    assert has_no_css?("img[src='#{@valid_url}']")
+
+    image = Image.create!(url: @valid_url, tag_list: 'one, two, three')
+    visit(image_path(image))
+    assert has_css?("img[src='#{@valid_url}']")
+    click_link('Delete')
+    dialog.accept
```
rmacklin: This is a stale reference as the alert being referred to by `dialog` here was already accepted on line 31. That means this is flakily passing. You'll want to reassign `dialog` first.

test/flows/images_flow_test.rb
```diff
@@ -29,6 +29,17 @@ class ImagesFlowTest < FlowTestCase
     click_link('Delete')
     dialog = page.driver.browser.switch_to.alert
     dialog.accept
+    assert page.has_current_path? images_path
+    assert page.has_content?('Image successfully deleted!')
+    assert has_no_css?("img[src='#{@valid_url}']")
+
+    image = Image.create!(url: @valid_url, tag_list: 'one, two, three')
```
rmacklin: It would be better to have two images _with different URLs_ before the first one is deleted, so that you can verify that the _correct_ image is deleted. We can also split testing image deletion to its own test case.

test/controllers/images_controller_test.rb
```diff
@@ -103,16 +103,14 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
-  test 'delete removes image successfully' do
+  test 'test removes image successfully' do
```
rmacklin: ❓

test/controllers/images_controller_test.rb
```diff
@@ -103,16 +103,14 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
     image = Image.create!(url: 'https://www.google.com/image2.png')
 
     assert_difference 'Image.count', -1 do
       delete image_path(image)
     end
-    assert_response :ok
-
-    response_hash = JSON.parse(@response.body)
-    assert_equal({ 'image_id' => image.id }, response_hash)
+    assert_response :redirect
```
rmacklin: You should always use `assert_redirected_to` so that you assert both that you were redirected as well as where you were redirected to.

test/flows/images_flow_test.rb
ragurney: 2f6bb1b

test/flows/images_flow_test.rb
ragurney: 1f27848

test/controllers/images_controller_test.rb
ragurney: 91e7330

test/controllers/images_controller_test.rb
ragurney: I have no idea why that changed. Fixed: 91e7330

test/flows/images_flow_test.rb
```diff
@@ -19,17 +19,39 @@ class ImagesFlowTest < FlowTestCase
     visit(images_path)
     assert (has_current_path? images_path), 'root path should exist'
     assert has_css?("img[src='#{@valid_url}']"), 'home page should list added image'
+  end
 
-    click_link('Delete')
+  test 'test delete functionality' do
```
rmacklin: We don't need to include "test" in the test name. Also, the test name for the first test is now inaccurate.

PR #38, opened 2016-09-14
Flow tests with page objects

### LINKED ISSUE #33: Refactor flow tests to use page objects
Reference material:
 - https://github.com/appfolio/ae_page_objects
 - https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/flow-tests-with-page-objects

The starting skeleton is on the `flowTestsWithPageObjects` branch.

ragurney:
Closes #33


test/flows/images_crud_test.rb
```diff
@@ -7,51 +7,60 @@ class ImagesCrudTest < FlowTestCase
...
+
+    image_url = 'https://media.giphy.com/media/rl0FOxdz7CcxO/giphy.gif'
     new_image_page.url.set(image_url)
 
     image_show_page = new_image_page.create_image!
-    assert_equal 'You have successfully added an image.', image_show_page.flash_message(:success)
+    assert_equal 'Url successfully saved!', image_show_page.flash_message(:success)
 
-    assert image_show_page.has_url?(image_url)
+    assert image_show_page.url?(image_url)
```
rmacklin: I'm not a fan of this change. `url?` is not a clear method name. We could change it to
```ruby
assert_equal image_url, image_show_page.image_url
```

test/page_objects/images/new_page.rb
```diff
@@ -1,12 +1,19 @@
...
+      path :new_image
+      path :images # if create fails
 
-      form_for :image #TODO
+      form_for :image do
+        element :url
+        element :tag_list
+      end
 
       def create_image!(url: nil, tags: nil)
```
rmacklin: It's better to have this match the button text: `upload_image!` (though the button is inaccurate, since we're not uploading it...`add_image!` would be more appropriate if the button said Add Image)

test/page_objects/images/show_page.rb
```diff
@@ -3,28 +3,29 @@ module Images
...
-        #TODO
+        node.click_on('Delete')
         yield node.driver.browser.switch_to.alert
       end
 
       def delete_and_confirm!
-        #TODO
+        node.click_on('Delete')
         alert = node.driver.browser.switch_to.alert
         alert.accept
```
rmacklin: You can refactor these three lines to `delete(&:accept)`

test/page_objects/images/index_page.rb
```diff
@@ -3,9 +3,26 @@ module Images
...
+          yield node.session.driver.browser.switch_to.alert
+          window.change_to(IndexPage)
+        end
+
+        def delete_image_and_confirm!
+          node.click_on('Delete')
+          alert = node.session.driver.browser.switch_to.alert
+          alert.accept
+          window.change_to(IndexPage)
         end
```
rmacklin: These methods are almost the same as those in `ShowPage`, with the exception being `delete_image!` has some issues and should instead be the same method as `ShowPage#delete`.

That said, we should DRY the methods. We can push them into `ImageCard`

test/flows/images_crud_test.rb
ragurney: ee9548e

test/page_objects/images/show_page.rb
ragurney: 44622b2

test/page_objects/images/index_page.rb
ragurney: e346235

test/page_objects/images/new_page.rb
ragurney: 89bbae5

test/flows/images_crud_test.rb
```diff
@@ -49,13 +49,13 @@ class ImagesCrudTest < FlowTestCase
...
 
-    image_show_page.delete do |confirm_dialog|
+    image_show_page.image.delete do |confirm_dialog|
       assert_equal 'Are you sure you want to delete this image?', confirm_dialog.text
       confirm_dialog.dismiss
     end
     assert_equal mind_blown_url, image_show_page.image_url
 
-    images_index_page = image_show_page.delete_and_confirm!
+    images_index_page = image_show_page.image.delete_and_confirm!
```
rmacklin: You can avoid these changes, and keep the cleaner/simpler interface if you use `delegate` to delegate these methods to the underlying `ImageCard` instance

ragurney: 85d1dd9

ragurney: 7d9375f

rmacklin:
Looks good. Let's squash the last commit away so that the history doesn't modify those lines to insert `.image` and then immediately undo that change.

PR #42, opened 2016-09-28
Image share with modal

### LINKED ISSUE #10: Share Image

ragurney:
Closes #10 
Image share with modal


config/initializers/assets.rb
```diff
@@ -9,3 +9,5 @@
 # Precompile additional assets.
 # application.js, application.scss, and all non-JS/CSS in app/assets folder are already added.
 # Rails.application.config.assets.precompile += %w( search.js )
+Rails.application.config.assets.precompile += %w( images/index/index.js )
+Rails.application.config.assets.precompile += %w( images/show/index.js )
```
lsq921224: consider combine these two lines to
`Rails.application.config.assets.precompile += %w( images/index/index.js images/show/index.js)`

app/assets/javascripts/images/share_image_modal.js
```diff
@@ -5,16 +5,40 @@ class ShareImageModal {
...
+    }
+  }
+
+  static generateFlashMessage(type, msg) {
+    if ($(`div.alert-${type}:contains(${msg})`).length <= 0) {
+      ShareImageModal.deleteFlashMessage();
+      $('.js-body-container').prepend(`<div class="alert alert-${type}">${msg}</div>`);
+    }
+  }
+
```
lsq921224: I don't think `deleteFlashMessage()` and `generateFlashMessage(type, msg)` should belong to `share_image_modal.js`, maybe move them to another js file like `set_flash_messages.js`?

test/flows/share_image_test.rb
```diff
@@ -33,6 +56,7 @@ class ShareImageTest < FlowTestCase
     assert_equal 'Email successfully sent!', images_index_page.flash_message(:success)
   end
 
+  # TODO: // Currently broken, will be fixed in subsequent commit
```
lsq921224: Is there anyway we can fix this test in this commit?

test/page_objects/images/show_page.rb
```diff
@@ -21,6 +21,14 @@ def go_back_to_index!
...
         window.change_to(IndexPage)
       end
+
+      def share_image
+        node.click_on('Share')
+        modal = element(locator: '#share_image_modal', is: ShareModal)
+        modal.wait_until_visible
+        yield modal
+        modal.wait_until_hidden
+      end
```
lsq921224: this may cause flakiness for  `test 'share nonexistent image'`, please see this comment https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50#discussion_r78908412

app/assets/javascripts/images/share_image_modal.js
```diff
@@ -34,10 +36,16 @@ class ShareImageModal {
       this.$shareFormImage.attr('src', imageUrl);
     });
 
-    $(document).on('ajax:success', '#share_image_modal', () => {
+    this.$modal.on('hide.bs.modal', event => {
+      this.clearForm();
```
lsq921224: We can add a flow test for this: `test the modal gets reset when being hidden`. Right now I don't think the form is reset to its original state:

see this:
http://recordit.co/xCefNSg4Bk

app/assets/javascripts/images/share_image_modal.js
```diff
@@ -34,10 +36,16 @@ class ShareImageModal {
...
+    });
+
+    $(document).on('ajax:success', '#share_image_modal', (e, xhr) => {
       this.clearForm();
       this.$modal.modal('hide');
       ShareImageModal.generateFlashMessage('success', 'Email successfully sent!');
+    }).on('ajax:error', '#share_image_modal', (e, xhr) => {
+      this.$shareForm.html(xhr.responseText)
```
lsq921224: > We should use a JSON protocol with a specific key for this rather than just relying on it being the response text.

app/views/images/_share_new.html.erb
```diff
@@ -1,23 +1,19 @@
...
-        </div>
+      <%= f.input :email_address,
+                  label: "Recipient's email",
+                  input_html: { placeholder: 'Enter email'} %>
+      <%= f.input :message,
+                  label: 'Message (optional)',
+                  as: :text,
+                  input_html: { placeholder: 'Enter message'} %>
+      <div class="card text-xs-center">
+        <%= image_tag image ? image.url : '', class: 'image-card image-card__figure' %>
```
lsq921224: could use `image.try(:url).to_s` here

app/views/images/_share_new_modal.html.erb
```diff
@@ -7,7 +7,7 @@
         </button>
         <h4 class="modal-title" id="share_image_modal_label">Modal title</h4>
       </div>
-      <%= render 'share_new', share_form: share_form %>
+        <%= render 'share_new', share_form: share_form, image: image %>
```
lsq921224: trivial: I think this extra space is not intended

test/flows/share_image_test.rb
```diff
@@ -14,13 +14,36 @@ class ShareImageTest < FlowTestCase
...
+    images_show_page.share_image do |modal|
+      assert_equal happening_url, modal.image_url
+      modal.email_address.set('valid@email.com')
+      modal.message.set('Message!')
+      modal.send_email
+    end
+    assert_equal 'Email successfully sent!', images_show_page.flash_message(:success)
+  end
 
+  # TODO: // Currently broken, will be fixed in subsequent commit
```
rmacklin: As mentioned in my previous commit, I think the subsequent commit being split out from this one doesn't make a whole lot of sense, and given that it caused you to add these comments, the split is definitely not atomic.

test/page_objects/images/index_page.rb
```diff
@@ -29,6 +29,13 @@ def clear_tag_filter!
         node.click_on('All Images')
         window.change_to(IndexPage)
       end
+
+      def share_image
```
rmacklin: As discussed on slack, this should be provided by `ImageCard`, not `IndexPage`

test/page_objects/images/image_card.rb
```diff
@@ -24,9 +24,8 @@ def delete_and_confirm!
         window.change_to(IndexPage)
       end
 
-      def open_share_page!
+      def open_share_modal
```
rmacklin: This method should return an instance of the modal, so it can be used for testing the 404 case (see also https://github.com/AppFolioOnboarding/image-sharer-ragurney/pull/42#discussion_r80969785).

test/page_objects/images/share_modal.rb
```diff
@@ -0,0 +1,18 @@
...
+      form_for :share_form do
+        element :email_address
+        element :message
+      end
+
+      def image_url
+        node.find('img')[:src]
+      end
+
+      def send_email
```
rmacklin: You should just call this `send` since that matches the button text, so it maps better to what a user would call this action.

app/assets/javascripts/images/share_image_modal.js
```diff
@@ -19,11 +16,16 @@ class ShareImageModal {
...
   static generateFlashMessage(type, msg) {
     if ($(`div.alert-${type}:contains(${msg})`).length <= 0) {
-      ShareImageModal.deleteFlashMessage();
+      ShareImageModal.clearFlashMessage();
       $('.js-body-container').prepend(`<div class="alert alert-${type}">${msg}</div>`);
     }
   }
 
+  clearForm() {
+    this.$shareForm.empty().append(this.cleanFormHTML);
```
rmacklin: This can just be `this.$shareForm.html(this.cleanFormHTML);`

app/assets/javascripts/images/share_image_modal.js
```diff
@@ -19,11 +16,16 @@ class ShareImageModal {
...
+    this.constructor('#share_image_modal');
```
rmacklin: Calling the constructor again for an already-constructed object is a bad pattern. As the constructor gains additional functionality, this could lead to bugs. In this case, it looks like it's only being done so that `this.$shareFormImage` will be reset. We could just remove that property and replace the one usage with `this.$shareForm.find('img')`.

app/controllers/images_controller.rb
```diff
@@ -34,8 +34,8 @@ def share_send
                                    url: @image.url).deliver_now
       head :ok
     else
-      # TODO: handle error
-      head :unprocessable_entity
+      render plain: render_to_string(partial: 'share_new', locals: { share_form: @share_form, image: @image }),
```
rmacklin: As mentioned in previous comments, it would be better to use a JSON response here, and if we renamed the form partial, this could be `render_to_string(partial: 'share_form', object: @share_form, locals: { image: @image })`

app/views/images/_share_new.html.erb
```diff
@@ -1,23 +1,19 @@
 <%= simple_form_for share_form, remote: true, url: ''  do |f| %>
-        <div class="modal-body">
-          <div class="row">
-            <div class="col-xs-6">
-              <%= f.error_notification  %>
+    <div class="modal-body">
+      <%= f.error_notification  %>
```
rmacklin: The removal of unneeded divs and the indentation fixing can be squashed out of this commit (45dc077e7f415cd40a4fbfc10f27ab0aac73f699).

test/flows/share_image_test.rb
```diff
@@ -36,27 +36,35 @@ class ShareImageTest < FlowTestCase
...
 
-    share_new_image_page.email_address.set('invalid_email')
-    share_new_image_page = share_new_image_page.share_image!.as_a(PageObjects::Images::ShareNewPage)
-    assert_equal 'Please review the problems below:', share_new_image_page.flash_message(:danger)
-    assert_equal 'is invalid', share_new_image_page.email_address.error_message
+    images_show_page.share_image do |modal|
+      assert_equal happening_url, modal.image_url
+      modal.send_email
+      assert_equal 'Please review the problems below:', images_show_page.flash_message(:danger)
+      assert_equal "can't be blank", modal.email_address.error_message
```
rmacklin: I like that you added testing submitting a blank recipient as well, but the diff and the history would be much clearer if you added that assertion to the existing flow test _prior_ to refactoring the flow test to support the image modal.

In general, if you're going to add anything to the flow test, try to do that first, so that the refactor-to-modal diff is as small and straightforward as possible.

test/flows/share_image_test.rb
```diff
@@ -65,11 +73,15 @@ class ShareImageTest < FlowTestCase
...
-    share_new_image_page.email_address.set('valid@email.com')
-    share_new_image_page.message.set('Some message')
-    images_index_page = share_new_image_page.share_image!
+
+    image_to_share.open_share_modal
+    images_index_page.share_image_redirect do |modal|
+      created_image.destroy
+      modal.email_address.set('valid@email.com')
+      modal.message.set('Some message')
+      images_index_page = modal.send_email_with_redirect!
```
rmacklin: `send_email_with_redirect!` isn't a method that should exist on the modal. This is one of the rare cases where there's no perfect solution using our page object model, so the recommendation is to use `AePageObjects.browser.current_window.change_to`. Peter can help you with this.

test/page_objects/images/index_page.rb
```diff
@@ -36,6 +36,12 @@ def share_image
         yield modal
         modal.wait_until_hidden
       end
+
+      def share_image_redirect
```
rmacklin: The page object interface should reflect how a user things about interacting with the application. A user isn't going to think "share image with redirect", they're just thinking "share image". That means this is not a good page object method. Talk to Peter about a good page object interface.

test/flows/share_image_test.rb
ragurney: :heavy_check_mark:

ragurney:
Superseded by #50

PR #45, opened 2016-10-03
Fix controller tests

rmacklin:


rmacklin:
@ragurney Please take a look at these changes that address issues with the existing controller tests.

PR #51, opened 2016-10-07
Authentication

### LINKED ISSUE #11: Authentication
#### As an application hoster I want to add user authentication to my application.

__Story__:

We have done a great job providing immediate value to our customers. At this
point they can both add and delete images, in addition to a few other features
that are either completed or in the works. Along with a growth in our web
service's popularity have come an increase in number of reports of missing
images. Upon investigation, it turns out that the majority of delete requests
are issued by only a handful of IP addresses. While we could blacklist those IP
addresses from the delete operation, it has always been our plan to only permit
the user who added an image to delete that image. To begin that effort we must
first add the concept of a user.

__Acceptance criteria__:
- [x] Users can sign up.
- [x] Users can log in after they've signed up.
- [x] Users can log out after logging in.
- [x] Users can choose to be remembered after logging in.<sup>†</sup>
- [x] Multiple browsers can concurrently hold permanent cookies for the same
  user.
- [x] Logged in users should not be able to register, nor log in again.

<sup>†</sup>: "Remembering" means that users will remain logged into the site
even after completely closing their web browser.

__Note__: Do not use Devise for this task.

__Discussion topics__:
- [ ] If we wanted to add more advanced features such as email activation and
  "forgot my password", should we continue to "roll our own" authentication?

__Reference material__:
- https://www.railstutorial.org/book/log_in_log_out

__Further reading__:
- https://github.com/plataformatec/devise

__Dependencies__:
- Delete Image

ragurney:
Closes #11

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,7 @@ gem 'jbuilder', '~> 2.5'
 # Use Redis adapter to run Action Cable in production
 # gem 'redis', '~> 3.0'
 # Use ActiveModel has_secure_password
-# gem 'bcrypt', '~> 3.1.7'
+gem 'bcrypt', '~> 3.1.7'
 gem 'simple_form', '3.2.1'
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,0 +1,5 @@
+class SessionsController < ApplicationController
+  def new
+    @session = Session.new
+  end
+end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -32,6 +32,7 @@ class SessionsController < ApplicationController
     user = User.find_by(email: session_params[:email].downcase)
     if user && user.authenticate(@session.password)
       log_in(user)
+      remember(user) if @session.remember_me == '1'
       redirect_to(root_path, flash: { success: 'Successfully signed in!' })
     else
       flash[:danger] = 'There was a problem with your username and/or password'
@@ -40,7 +41,16 @@ class SessionsController < ApplicationController
   end
 
   def log_out
+    forget
     session.delete(:user_id)
     @current_user = nil
+  end
+
+  def remember(user)
+    cookies.permanent.signed[:user_id] = user.id
+  end
+
+  def forget
+    cookies.delete(:user_id)
   end
 end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,5 +1,33 @@
 class SessionsController < ApplicationController
+  include Loginable
+
   def new
     @session = Session.new
+  end
+
+  def create
+    @session = Session.new(session_params)
+    if @session.valid?
+      authenticate_and_login
+    else
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def session_params
+    params.require(:session).permit(:email, :password, :remember_me)
+  end
+
+  def authenticate_and_login
+    user = User.find_by(email: session_params[:email].downcase)
+    if user && user.authenticate(@session.password)
+      log_in(user)
+      redirect_to(root_path, flash: { success: 'Successfully signed in!' })
+    else
+      flash[:danger] = 'There was a problem with your username and/or password'
+      render :new, status: :bad_request
+    end
   end
 end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,6 +1,8 @@
 class SessionsController < ApplicationController
   include Loginable
 
+  before_action :redirect_on_authenticated, only: [:new, :create]
+
   def new
     @session = Session.new
   end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -16,6 +16,12 @@ class SessionsController < ApplicationController
     end
   end
 
+  def destroy
+    log_out if logged_in?
+    flash[:success] = 'Successfully logged out'
+    redirect_to root_path
+  end
+
   private
 
   def session_params
@@ -31,5 +37,10 @@ class SessionsController < ApplicationController
       flash[:danger] = 'There was a problem with your username and/or password'
       render :new, status: :bad_request
     end
+  end
+
+  def log_out
+    session.delete(:user_id)
+    @current_user = nil
   end
 end
--- a/app/forms/session.rb
+++ b/app/forms/session.rb
@@ -1,0 +1,7 @@
+class Session
+  include ActiveModel::Model
+  validates :email, presence: true
+  validates :password, presence: true
+
+  attr_accessor :email, :password
+end
--- a/app/forms/session.rb
+++ b/app/forms/session.rb
@@ -3,5 +3,5 @@ class Session
   validates :email, presence: true
   validates :password, presence: true
 
-  attr_accessor :email, :password
+  attr_accessor :email, :password, :remember_me
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,6 +28,9 @@ module Base
     config.browserify_rails.commandline_options = '-t babelify'
 
     # Custom directories
-    config.autoload_paths += Dir["#{config.root}/business_logic/**/"]
+    config.autoload_paths += %W(
+      #{Rails.root}/business_logic
+      #{Rails.root}/lib
+    )
   end
 end
--- a/lib/loginable.rb
+++ b/lib/loginable.rb
@@ -4,7 +4,8 @@ module Loginable
   end
 
   def current_user
-    @current_user ||= User.find_by(id: session[:user_id])
+    user_id = session[:user_id] || cookies[:user_id]
+    @current_user ||= User.find_by(id: user_id)
   end
 
   def logged_in?
--- a/lib/loginable.rb
+++ b/lib/loginable.rb
@@ -1,0 +1,13 @@
+module Loginable
+  def log_in(user)
+    session[:user_id] = user.id
+  end
+
+  def current_user
+    @current_user ||= User.find_by(id: session[:user_id])
+  end
+
+  def logged_in?
+    current_user.present?
+  end
+end
--- a/lib/loginable.rb
+++ b/lib/loginable.rb
@@ -10,4 +10,8 @@ module Loginable
   def logged_in?
     current_user.present?
   end
+
+  def redirect_on_authenticated
+    redirect_to root_path, flash: { danger: 'Oops! You are already logged in!' } if logged_in?
+  end
 end
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -8,4 +8,81 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_select 'h1', 'Log In'
     assert_select 'form#new_session', 1
   end
+
+  test 'should show inline errors when email field is blank' do
+    post sessions_path,
+         params: {
+           session: {
+             email: '',
+             password: '[FILTERED]'
+           }
+         }
+
+    assert_response :unprocessable_entity
+    assert_select '.session_email .text-muted', text: "can't be blank"
+  end
+
+  test 'should re-render with flash error message if user is not found' do
+    post sessions_path,
+         params: {
+           session: {
+             email: 'stuff',
+             password: '[FILTERED]'
+           }
+         }
+
+    assert_response :bad_request
+    assert_equal 'There was a problem with your username and/or password', flash[:danger]
+    assert_select 'input.email[value=?]', 'stuff'
+  end
+
+  test 'should re-render with flash error message if password does not match user' do
+    User.create!(email: 'valid@email.com', password: 'password123')
+    post sessions_path,
+         params: {
+           session: {
+             email: 'valid@email.com',
+             password: 'invalid'
+           }
+         }
+
+    assert_response :bad_request
+    assert_equal 'There was a problem with your username and/or password', flash[:danger]
+    assert_select 'input.email[value=?]', 'valid@email.com'
+  end
+
+  test 'should redirect to index upon successful login' do
+    user = User.create!(email: 'valid@email.com', password: 'password123')
+    login_as(user)
+
+    assert_equal 'Successfully signed in!', flash[:success]
+  end
+
+  test 'should find correct user even with uppercase email and log in successfully' do
+    User.create!(email: 'valid@email.com', password: 'password123')
+    post sessions_path,
+         params: {
+           session: {
+             email: 'ValID@email.com',
+             password: 'password123'
+           }
+         }
+
+    assert_redirected_to root_path
+    assert_equal 'Successfully signed in!', flash[:success]
+  end
+
+  private
+
+  def login_as(user)
+    post sessions_path,
+         params: {
+           session: {
+             email: user.email,
+             password: user.password
+           }
+         }
+
+    assert_redirected_to root_path
+  end
 end
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -72,6 +72,19 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Successfully signed in!', flash[:success]
   end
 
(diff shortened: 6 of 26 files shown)


test/flows/user_sign_up_test.rb
```diff
@@ -0,0 +1,27 @@
...
+  test 'user sign up with invalid information re-render back to sign up' do
+    users_new_page = PageObjects::Users::NewPage.visit
+
+    users_new_page.email.set('valid@email.com')
+    users_new_page = users_new_page.add_user_with_error!
+
+    assert_equal 'valid@email.com', users_new_page.email.value
+    assert_equal 'Please review the problems below:', users_new_page.flash_message(:danger)
+  end
+end
```
bboe: Similar comments. Think about a new user to your application. What are they going to do in order to perform an action on your site? Think about this in the context of your `new user flow`.

Similarly, you might have a already-existing user flow, where they submit and delete an image. These flow tests should not, however, test isolated actions (e.g., login, register, submit).

test/page_objects/users/new_page.rb
```diff
@@ -0,0 +1,29 @@
...
+      path :new_user
+      path :users # if create fails
+
+      form_for :user do
+        element :email
+        element :password
+        element :password_confirmation
+      end
+
+      def click_sign_up!
```
bboe: From the sign up page, why would I click on sign up again?

app/controllers/concerns/sessionable.rb
```diff
@@ -16,4 +16,8 @@ def current_user
   def logged_in?
     !current_user.nil?
   end
+  
+  def redirect_on_authenticated
+    redirect_to root_path if logged_in?
```
bboe: Is there a message we could show to the users in these cases? It's always nice to provide them with some reason why they were redirected.

test/controllers/sessions_controller_test.rb
```diff
@@ -48,4 +48,15 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to root_path
     assert_equal 'Successfully signed in!', flash[:success]
   end
+
+  test 'should not allow access to log in page once logged in' do
+    User.create!(email: 'valid@email.com', password: 'password123')
+    post login_path, params: { session: { email: 'valid@email.com', password: 'password123' } }
```
bboe: How can you mimic a logged in user without having to issue two requests?

test/controllers/users_controller_test.rb
```diff
@@ -34,4 +34,15 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     assert_response :unprocessable_entity
     assert_select '.user_email .text-help', 'has already been taken'
   end
+
+  test 'should not allow access to sign up page once logged in' do
+    User.create!(email: 'valid@email.com', password: 'password123')
+    post login_path, params: { session: { email: 'valid@email.com', password: 'password123' } }
```
bboe: How can you mimic a logged in user without having to issue two requests?

test/controllers/sessions_controller_test.rb
```diff
@@ -48,4 +48,15 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
...
+
+    get login_path
+    assert_redirected_to root_path
+
+    post login_path, params: { session: { email: 'valid@email.com', password: 'password123' } }
```
bboe: Should get and post be separate tests?

test/controllers/users_controller_test.rb
```diff
@@ -34,4 +34,15 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
...
+
+    get new_user_path
+    assert_redirected_to root_path
+
+    post users_path(user: { email: 'valid@email.com', password: '123password' })
```
bboe: Should this be a separate test?

app/controllers/concerns/sessionable.rb
```diff
@@ -16,4 +16,8 @@ def current_user
   def logged_in?
```
bboe: > Do not allow users already logged in to sign up or log in again

good work ⭐️

test/controllers/sessions_controller_test.rb
```diff
@@ -49,6 +49,14 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Successfully signed in!', flash[:success]
   end
 
+  test 'successful log in with remember me selected' do
+    User.create!(email: 'valid@email.com', password: 'password123')
+    post login_path, params: { session: { email: 'valid@email.com', password: 'password123', remember_me: 1 } }
+
+    assert_redirected_to root_path
```
bboe: How can you assert that remember me worked as expected and that the remember me side effect doesn't occur with when "remember me" is not checked?

Gemfile
```diff
@@ -25,6 +25,7 @@ gem 'simple_form', '3.2.1'
 # gem 'capistrano-rails', group: :development
 gem 'acts-as-taggable-on', '~> 4.0'
 gem 'bootstrap', '~> 4.0.0.alpha3'
+gem 'bcrypt', '~> 3.1'
```
ragurney: Addressed b718c7ed993ce492110186644904cc0ea14832db

test/controllers/users_controller_test.rb
ragurney: Addressed 4a156f31f1d616a373e09a48a452c4495d4076b7

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,4 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+  include Sessionable
```
ragurney: Right now, yes I believe so. This is mainly due to the navigation bar, which uses the `logged_in?` concern method. Because this navbar is in every controller's view, then every controller must be sessionable so that those views can call `logged_in?`.

app/controllers/sessions_controller.rb
```diff
@@ -2,4 +2,29 @@ class SessionsController < ApplicationController
   def new
     @session = Session.new
   end
+
+  def create
+    @session = Session.new(session_params)
+    if @session.valid?
+      handle_authenticate_and_login(User.find_by(email: @session.email.downcase))
+    else
+      render :new
```
ragurney: D'oh! I'll go ahead and add the status.

app/controllers/concerns/sessionable.rb
```diff
@@ -0,0 +1,49 @@
...
+      user = User.find_by(id: user_id)
+      if user && user.authenticated?(cookies[:remember_token])
+        log_in user
+        @current_user = user
+      end
+    end
+  end
+
+  def logged_in?
+    current_user
```
rmacklin: Methods ending in `?` should return a boolean. You could do `current_user.present?`

test/page_objects/images/index_page.rb
```diff
@@ -29,6 +29,16 @@ def clear_tag_filter!
         node.click_on('All Images')
         window.change_to(IndexPage)
       end
+
+      def click_log_in!
+        node.click_on('Log in')
+        window.change_to(Sessions::NewPage)
+      end
+
+      def click_log_out!
```
rmacklin: This method should just be called `log_out!`

test/page_objects/images/index_page.rb
```diff
@@ -29,6 +29,16 @@ def clear_tag_filter!
         node.click_on('All Images')
         window.change_to(IndexPage)
       end
+
+      def click_log_in!
```
rmacklin: This method should be called either `log_in!` or `open_login_page!`

app/controllers/application_controller.rb
rmacklin: Generally, it's bad to share more than you need to share. The `log_out`, `forget` and `remember` methods are only needed in the sessions controller so they can/should be inlined there. The other methods could be inside a "Loginable" module. Rather than including `Loginable` in `ApplicationController`, you can include it only in `SessionsController` and `UsersController` (the two controllers that need it), as well as `Application`_`Helper`_, the place for view helper methods needed by the application layout (i.e. `logged_in?`). By including it in the helper, you would turn `Loginable` into a plain module (not including `ActiveSupport::Concern`).

test/controllers/sessions_controller_test.rb
```diff
@@ -8,4 +8,68 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
...
+         }
+
+    assert_response :bad_request
+    assert_equal 'There was a problem with your username and/or password', flash[:danger]
+    assert_select 'input.email[value=?]', 'valid@email.com'
+  end
+
+  test 'should redirect to index upon successful login' do
+    user = User.create!(email: 'valid@email.com', password: 'password123')
+    login_as(user)
```
bboe: Where is this coming from?

test/controllers/sessions_controller_test.rb
```diff
@@ -73,6 +73,16 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Successfully signed in!', flash[:success]
   end
 
+  test 'log out successfully, redirect to index page and show proper success message' do
+    user = User.create!(email: 'valid@email.com', password: 'password123')
+    login_as(user)
+
+    assert_equal user.id, session[:user_id]
+    delete session_path
+    assert_redirected_to root_path
```
bboe: Is there a way to verify that the session cookie no longer exists? That's one of the side-effects that would be nice to test for.

If it cannot be done through checking for a set-cookie header that clears the cookie, perhaps we should employ an additional request and verify that the application is in its logged out state.

test/flows/sessions_test.rb
```diff
@@ -0,0 +1,44 @@
...
+    ).as_a(PageObjects::Sessions::NewPage)
+
+    assert_equal "can't be blank", sessions_new_page.email.error_message
+
+    sessions_new_page = sessions_new_page.log_in!(
+      email: @user.email,
+      password: 'Not the password'
+    ).as_a(PageObjects::Sessions::NewPage)
+
+    assert_equal 'There was a problem with your username and/or password',
```
bboe: I don't think we need to test two failure cases in the flow tests. You have that already in the controller tests. A single instance of red and then green should suffice.

test/controllers/sessions_controller_test.rb
```diff
@@ -72,4 +72,40 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
...
+  private
+
+  def login_as(user)
+    post sessions_path,
+         params: {
+           session: {
+             email: user.email,
+             password: user.password
+           }
+         }
```
bboe: Should this make an assertion that it was successful?

test/controllers/users_controller_test.rb
```diff
@@ -26,12 +26,33 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     assert_select '.user_email .text-help', 'is an invalid email address'
   end
 
-  test 'create invalid user with already taken email' do
```
bboe: Why is this test being removed?

test/controllers/users_controller_test.rb
```diff
@@ -26,12 +26,33 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
...
+    login_as(user)
+
+    post users_path(user: { email: 'valid@email.com', password: '123password' })
+    assert_redirected_to root_path
+    assert_equal 'Oops! You are already logged in!', flash[:danger]
+  end
+
+  private
+
+  def login_as(user)
```
bboe: Can we dry private method up?

test/lib/loginable_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'test_helper'
+
+class LoginableTest < ActiveSupport::TestCase
+  class LoginableTestController < ActionController::Base
+    include Loginable
+  end
+
+  def setup
+    @controller = LoginableTestController.new
```
bboe: Does anything seem strange about the order of these statements?

lib/loginable.rb
```diff
@@ -4,7 +4,9 @@ def log_in(user)
   end
 
   def current_user
-    @current_user ||= User.find_by(id: session[:user_id])
+    user_id = session[:user_id] || cookies.signed[:user_id]
```
bboe: Do you need to used `signed` in order to access the value of `:user_id` from cookies?

test/controllers/sessions_controller_test.rb
```diff
@@ -118,4 +138,15 @@ def login_as(user)
            }
          }
   end
+
+  def login_and_remember_as(user)
```
bboe: Could `remember_as` be a keyword argument to `login_as` that defaults to false?

test/controllers/sessions_controller_test.rb
```diff
@@ -73,6 +73,15 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Successfully signed in!', flash[:success]
   end
 
+  test 'successful log in with remember me selected' do
+    user = User.create!(email: 'valid@email.com', password: 'password123')
+    login_and_remember_as(user)
+
+    assert_redirected_to root_path
+    assert_equal 'Successfully signed in!', flash[:success]
+    assert cookies['user_id']
```
bboe: What do you really want to assert here? Using a naked assert or refute doesn't provide much context when it fails.

test/controllers/sessions_controller_test.rb
```diff
@@ -83,6 +92,17 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Successfully logged out', flash[:success]
   end
 
+  test 'log out after remember me log in deletes user id in cookie' do
+    user = User.create!(email: 'valid@email.com', password: 'password123')
+    login_and_remember_as(user)
+    assert cookies['user_id']
```
bboe: Should this assertion be part of `login_and_remember_as` (or it's keyword-argument form)?

bboe:
Looks like this branch should also be rebased.

image

app/forms/session.rb
```diff
@@ -0,0 +1,7 @@
+class Session
+  include ActiveModel::Model
+  validates :email, presence: true
+  validates :password, presence: true
+
+  attr_accessor :email, :password, :remember_me
```
ragurney: I definitely see your point and I agree. I think it makes sense then to squash this into the `Implement new session 'login' action` commit.

config/application.rb
```diff
@@ -28,6 +28,9 @@ class Application < Rails::Application
     config.browserify_rails.commandline_options = '-t babelify'
 
     # Custom directories
-    config.autoload_paths += Dir["#{config.root}/business_logic/**/"]
+    config.autoload_paths += %W(
+      #{Rails.root}/business_logic
```
bboe: While I don't know if it needs to be fixed, this line is different than what was there before.

test/controllers/users_controller_test.rb
```diff
@@ -25,4 +25,34 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
...
+    assert_redirected_to root_path
+    assert_equal 'Oops! You are already logged in!', flash[:danger]
+  end
+
+  private
+
+  def login_valid_user
+    user = User.create!(email: 'valid@email.com', password: 'password123')
+
+    post sessions_path,
```
bboe: Do you not have access to login_as here?

test/controllers/sessions_controller_test.rb
```diff
@@ -72,6 +72,19 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'Successfully signed in!', flash[:success]
   end
 
+  test 'log out successfully, redirect to index page and show proper success message' do
```
bboe: What happens if you logout when already logged out?

lib/loginable.rb
```diff
@@ -4,7 +4,9 @@ def log_in(user)
   end
 
   def current_user
-    @current_user ||= User.find_by(id: session[:user_id])
+    user_id = session[:user_id] || cookies[:user_id]
+    log_in @current_user if @current_user ||= User.find_by(id: user_id)
```
bboe: Why is this calling `log_in` now when it didn't before?

config/application.rb
ragurney: I think it's enough to just specify the directory, which is why I made this change. _Technically_ it should have been a separate PR, but I thought I might be able to get away with it in this commit 😬

lib/loginable.rb
ragurney: Good question, that definitely shouldn't be happening. I've removed it.

PR #53, opened 2016-10-10
Required tag association

### LINKED ISSUE #12: Required Tag Association
#### As a user I want all images to have at least one tag.

__Story__:

The tag feature has been a hit. Our users love being able to filter images by
cats, SR-71s, and "AppFolio tshirts" to name a few. In fact, we have discovered
that images without any tags receive significantly less views than those with
tags. As a result, we have decided to require all images to have at least one
tag.

__Acceptance criteria__:
- [ ] I cannot save an image without adding at least one tag.
- [ ] An error message is associated with the appropriate input field when a
tag is not provided.
- [ ] All images have at least one tag.

__Discussion Topic__:
- [ ] What do you do about existing images that have no tags?

__Dependencies__:
- Image Tags

ragurney:
Closes #12


app/models/image.rb
```diff
@@ -3,4 +3,11 @@
 class Image < ActiveRecord::Base
   acts_as_taggable
   validates :url, presence: true, url: { allow_blank: true, message: 'must be a valid url' }
+  validate :validate_has_at_least_one_tag
```
pololee: Could we just use `validates :tag_list, presence: true, { message: 'must have at least one tag' }`?

As you can see here http://guides.rubyonrails.org/active_record_validations.html#presence,

> it uses the blank? method

which covers `empty?` http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F

ragurney: fe2df00

PR #54, opened 2016-10-20
Modify tag list

### LINKED ISSUE #13: Modify Tag List
#### As a user I want to edit the tag list for an image.

__Story__:

Image tags are awesome -- our user's love them! Tags help us make sense of the
plethora of images added to our system. But how annoying is it when our users
make spelling mistakes in their image tags? Super annoying! How awesome, and
useful would it be to allow for tags to be edited after an image is added?
Super awesome, and useful!

__Acceptance criteria__:
- [x] There is a link to edit the tags for an image.
- [x] The image whose tags are to be modified is displayed on the edit page.
- [x] On the edit page, I can change the tags to any valid tag list.
- [x] After saving the edit page with valid tags, my changes are visible
  throughout the rest of the application.

__Discussion__:
- [x] Should an image's URL be editable?

__Dependencies__:
- Image Tags

ragurney:
Closes #13

### DIFF
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -1,5 +1,5 @@
 class ImagesController < ApplicationController
-  before_action :find_image_or_redirect, only: [:show, :destroy, :edit]
+  before_action :find_image_or_redirect, only: [:show, :destroy, :edit, :update]
   before_action :find_image_or_head_not_found, only: :share
 
   def index
@@ -21,6 +21,15 @@ class ImagesController < ApplicationController
   end
 
   def edit
+  end
+
+  def update
+    if @image.update(image_update_params)
+      flash[:success] = 'Tags successfully updated'
+      redirect_to image_path(@image)
+    else
+      render :edit, status: :unprocessable_entity
+    end
   end
 
   def show
@@ -46,6 +55,10 @@ class ImagesController < ApplicationController
 
   def image_params
     params.require(:image).permit(:url, :tag_list)
+  end
+
+  def image_update_params
+    params.require(:image).permit(:tag_list)
   end
 
   def email_params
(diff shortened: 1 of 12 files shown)


app/controllers/images_controller.rb
```diff
@@ -20,6 +20,18 @@ def create
     end
   end
 
+  def edit
+  end
+
+  def update
+    if @image.update(image_update_params)
+      flash[:success] = 'Tags successfully updated'
+      redirect_to images_path
```
weixing2014: Minor, but if I were a user I might want to view the show page after updating image tags in order to make sure the tags are successfully updated.
Then I might `redirect_to image_path(@image)`.

test/page_objects/images/image_card.rb
```diff
@@ -36,6 +36,11 @@ def share
         yield modal
         modal.wait_until_hidden
       end
+
+      def edit_tags
```
weixing2014: You might want to add `!` after `edit_tags` since the method returns a new page object.

app/views/images/edit.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1 class="text-xs-center">Edit Tags</h1>
+
+<div class="row">
+  <div class="col-xs-6">
+    <%= simple_form_for @image do |f| %>
+        <%= f.error_notification  %>
+
+        <%= image_tag @image.url, class: 'text-xs-center card image-card', style: 'max-width: 400px;' %>
```
weixing2014: Maybe you want to create a new css class in `stylesheet` and put `max-width: 400px;` there. http://stackoverflow.com/questions/3142710/inline-styles-vs-classes

ragurney:
I've updated the code to address your suggestions.

PR #55, opened 2016-10-27
Authorization

ragurney:
Closes #14

PR #57, opened 2016-11-01
Like image

### LINKED ISSUE #15: Like Image
#### As a logged in user I want to like an image.

**Story**:

What modern web service is complete without the ability to "like" something?
Well, many, but this is an image sharing site and our users demand to know how
many people have liked their image. Let's provide it to them!

**Acceptance criteria**:
- [ ] Each image on the index page has a "like" button.
- [ ] The button indicates how many users have liked that image.
- [ ] When you like an image the button state changes to a "liked" state.
- [ ] When you like an image the count of users who have liked that image is
  updated.
- [ ] If you have already liked an image, clicking the button "unlikes" the
  image.
- [ ] Liking/unliking is done via AJAX.

**Note**: Do not use an external gem to implement liking. You already have
experience using gem-backed models from acts_as_taggable_on, and since there
won't always be a gem to rely on, we want you to get experience implementing
more advanced models and relations on your own.

**Discussion topics**:
- [ ] What can we do to cache the likes count?
- [ ] How can we make our images relation for the index page include whether
  the image was liked by the current user?
- [ ] How can we develop the front end and back end separately?
- [ ] How can we reuse templates on the server and client?

**Reference material**:
- http://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache
- http://guides.rubyonrails.org/active_record_querying.html#scopes
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template

**Dependencies**:
- Authorization

ragurney:
Closes #15 

~~Must merge #55 first.~~

Relevant work for this PR starts at commit `Implement Like model and migration` (3a6617f)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,8 @@ end
 gem 'validate_url'
 gem 'email_validator'
 gem 'pundit'
+gem 'font-awesome-rails'
+gem 'fetch-rails'
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
--- a/app/assets/javascripts/images/like_button.js
+++ b/app/assets/javascripts/images/like_button.js
@@ -1,0 +1,53 @@
+import setFlashMessage from './set_flash_message'
+
+class LikeButton {
+  constructor(element) {
+    this.element = element;
+    this.attachEventHandlers();
+
+    const tokenMetaTag = document.querySelector('meta[name="csrf-token"]');
+    this.csrfToken = tokenMetaTag && tokenMetaTag.content;
+  }
+
+  updateLikeBtn({likeCount}) {
+    const likeCountField = this.element.getElementsByClassName('js-like-count')[0];
+
+    this.element.blur();
+    likeCountField.innerHTML = likeCount;
+    this.element.classList.toggle('image-card__liked')
+  }
+
+  attachEventHandlers() {
+    this.element.addEventListener('click', e => {
+      e.preventDefault();
+      fetch( this.element.href, {
+        credentials: 'include',
+        method: 'post',
+        headers: {
+          'X-CSRF-Token': this.csrfToken,
+          "Accept": "application/json"
+        }
+      })
+        .then(response => {
+          if (response.ok) {
+            return response.json();
+          }
+          else if(response.status == 404) {
+            window.location.replace('/');
+          }
+          else if(response.status == 401) {
+            setFlashMessage('danger', 'You must log in to like images!');
+          }
+          else {
+            alert("You've encountered an unsupported error. Please try again.");
+          }
+        })
+        .then( data => {
+          (data != null) && this.updateLikeBtn(data)
+        })
+        .catch( e => alert('Something went terribly wrong. Please try again.'));
+    });
+  }
+}
+
+export default LikeButton;
--- a/app/assets/javascripts/images/like_button_driver.js
+++ b/app/assets/javascripts/images/like_button_driver.js
@@ -1,0 +1,5 @@
+import LikeButton from './like_button';
+
+const elements = document.getElementsByClassName('js-like-btn');
+
+[].forEach.call(elements, element => new LikeButton(element));
--- a/app/assets/stylesheets/_image-card.scss
+++ b/app/assets/stylesheets/_image-card.scss
@@ -17,3 +17,13 @@
   font-size: 1em;
   margin-bottom: 2px;
 }
+
+.image-card__likebutton {
+  float: right;
+
+  &.image-card__liked {
+    background-color: #c1d9ff;
+  }
+}
+
+
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -1,8 +1,8 @@
 class ImagesController < ApplicationController
   include Loginable
   before_action :find_image_or_redirect, only: [:show, :destroy, :edit, :update]
-  before_action :find_image_or_head_not_found, only: :share
-  before_action -> { save_previous_path(request) }, except: :share
+  before_action :find_image_or_head_not_found, only: [:share, :like]
+  before_action -> { save_previous_path(request) }, except: [:share, :like]
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
@@ -44,17 +44,25 @@ class ImagesController < ApplicationController
   def share
     @share_form = ShareForm.new(email_params)
 
-    if @share_form.valid?
-      respond_to_valid_share_form
-    else
-      respond_to_invalid_share_form
-    end
+    @share_form.valid? ? respond_to_valid_share_form : respond_to_invalid_share_form
   end
 
   def destroy
     authorize @image
     @image.destroy
     redirect_to images_path, flash: { success: 'Image successfully deleted!' }
+  end
+
+  def like
+    authorize @image
+
+    if @image.liked_by?(current_user)
+      Like.find_by(image: @image, user: current_user).destroy
+    else
+      Like.create!(image: @image, user: current_user)
+    end
+
+    render json: { likeCount: @image.reload.likes_count }, status: :ok
   end
 
   private
@@ -104,6 +112,11 @@ class ImagesController < ApplicationController
   end
 
   def user_not_authorized
-    redirect_to new_session_path, flash: { danger: 'You must log in before accessing that page!' }
+    respond_to do |format|
+      format.json { head :unauthorized }
+      format.html do
+        redirect_to new_session_path, flash: { danger: 'You must log in before accessing that page!' }
+      end
+    end
   end
 end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -2,7 +2,13 @@ require 'uri'
 
 class Image < ActiveRecord::Base
   belongs_to :user
+  has_many :likes, dependent: :destroy
   acts_as_taggable
+
   validates :url, presence: true, url: { allow_blank: true, message: 'must be a valid url' }
   validates :tag_list, presence: { message: 'must have at least one tag' }
+
+  def liked_by?(current_user)
+    current_user.nil? ? false : Like.exists?(image: self, user: current_user)
+  end
 end
--- a/app/views/images/_image_card.html.erb
+++ b/app/views/images/_image_card.html.erb
@@ -23,5 +23,6 @@
                   class: "js-delete-image image-card__btn btn btn-danger",
                   'data-confirm': 'Are you sure you want to delete this image?' if policy(image).destroy? %>
     </div>
+    <%= render 'like_button', image: image, liked: image.liked_by?(current_user) %>
   </div>
 </div>
--- a/db/migrate/20161028173540_add_likes_table_and_add_like_to_image_and_user.rb
+++ b/db/migrate/20161028173540_add_likes_table_and_add_like_to_image_and_user.rb
@@ -1,0 +1,10 @@
+class AddLikesTableAndAddLikeToImageAndUser < ActiveRecord::Migration[5.0]
+  def change
+    create_table :likes do |t|
+      t.references :image, index: true, foreign_key: true, null: false
+      t.references :user, index: true, foreign_key: true, null: false
+    end
+
+    add_column :images, :likes_count, :integer, null: false, default: 0
+  end
+end
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -289,6 +289,23 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'The image you were looking for does not exist', flash[:danger]
   end
 
+  test 'Like image buttons should all use the unliked button class if not liked by current user' do
+    create_images
+
+    get images_path
+
+    assert_select 'a.image-card__likebutton', 4
+  end
+
+  test 'Like image buttons that were liked by current user should all use the liked button class' do
+    images = create_images
+    Like.create!(user: @user, image: images[0])
+
+    get images_path
+
+    assert_select 'a.image-card__likebutton', 4
+    assert_select 'a.image-card__likebutton.image-card__liked', 1
+  end
 
   test 'Liking an image should create a new Like record' do
     image = Image.create!(url: 'https://valid.com', tag_list: 'tag', user_id: @user.id)
--- a/test/controllers/images_controller_test.rb
+++ b/test/controllers/images_controller_test.rb
@@ -289,6 +289,37 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'The image you were looking for does not exist', flash[:danger]
   end
 
+
+  test 'Liking an image should create a new Like record' do
+    image = Image.create!(url: 'https://valid.com', tag_list: 'tag', user_id: @user.id)
+    assert_difference 'Like.count', 1 do
+      post like_image_path(image.id),
+           xhr: true
+    end
+    assert_response :ok
+  end
+
+  test 'Liking an image that user has already liked should decrement like count' do
+    image = Image.create!(url: 'https://valid.com', tag_list: 'tag', user_id: @user.id)
+    Like.create!(image: image, user: @user)
+
+    assert_difference 'Like.count', -1 do
+      post like_image_path(image.id),
+           xhr: true
+    end
+
+    assert_response :ok
+  end
+
+  test 'Like a non-existent image should redirect to image page and show flash message' do
+    assert_difference 'Like.count', 0 do
+      post like_image_path(-1),
+           xhr: true
+    end
+    assert_response :not_found
+    assert_equal 'The image you were looking for does not exist', flash[:danger]
+  end
+
   private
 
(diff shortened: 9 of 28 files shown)


db/migrate/20161028173540_add_likes_table_and_add_like_to_image_and_user.rb
```diff
@@ -0,0 +1,10 @@
+class AddLikesTableAndAddLikeToImageAndUser < ActiveRecord::Migration[5.0]
+  def change
+    create_table :likes do |t|
+      t.references :image, index: true, foreign_key: true, null: false
+      t.references :user, index: true, foreign_key: true, null: false
+    end
+
+    add_column :images, :likes_count, :integer, default: 0
```
tlconnor: What would happen here if I set `likes_count` to null?

test/models/like_test.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    like[:user_id] = @user.id
+    assert_predicate like, :valid?
+  end
+
+  test 'an image like count should be updated with each like created/destroyed' do
+    like = Like.new(user: @user, image: @image)
+
+    like.save
+    assert_equal 1, @image.likes_count
```
tlconnor: When you are performing an action that changes a count by some amount its cleaner to do something like:

```
assert_difference @image, :likes_count, 1 do
  like.save
end
```

test/models/like_test.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    like.destroy
```
tlconnor: Same thing as above (except assert a difference of `-1`)

test/models/like_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      { email: 'valid@email.com', password: 'password123' },
+      { email: 'valid1@email.com', password: 'password123' }
+    ])
+    image = Image.create!(url: 'https://valid.com/p.png', tag_list: 'tag', user_id: users[0].id)
+
+    like = Like.new(image: image, user: users[0])
+    assert_predicate like, :save
+
+    like = Like.new(image: image, user: users[0])
+    refute_predicate like, :save
```
tlconnor: You've tested that validation failed; could you also then check that it failed for the right reason? (i.e check `like.errors`)

test/controllers/images_controller_test.rb
```diff
@@ -293,6 +293,39 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'Liking an image that user has already liked should decrement like count' do
+    image = Image.create!(url: 'https://valid.com', tag_list: 'tag', user_id: @user.id)
+    Like.create!(image: image, user: @user)
+
+    assert_difference 'Like.count', -1 do
+      post like_image_path(image.id),
+           xhr: true
+    end
+
+    assert_equal 0, Like.count
```
tlconnor: You're already making an assertion about `Like.count` above so this line is probably unnecessary.

app/controllers/images_controller.rb
```diff
@@ -56,6 +56,17 @@ def destroy
...
 
+  def like
+    authorize @image
+
+    liked = Like.find_by(image: @image, user: current_user)
+    count = @image.likes_count
+
+    liked ? liked.destroy : Like.create!(image: @image, user: current_user)
+
+    render json: { likeCount: count, liked: liked.present?.to_s }, status: :ok
```
tlconnor: Any reason why `liked.present?` needs to be converted to a string?

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,44 @@
+import setFlashMessage from './set_flash_message'
+
+class LikeButton {
+  constructor(likeBtnClass) {
```
tlconnor: I find the name `likeBtnClass` to be a bit ambiguous.  The term `class` could mean a javascript `class`, or it could refer to CSS.  In this case it is CSS, so perhaps a clearer name would be something like `selector` or `locator`.

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,44 @@
...
+    this.likeBtnClass = likeBtnClass;
+  }
+
+  updateLikeBtn({likeCount, liked}, $likeBtn) {
+    var $likeCountField = $likeBtn.find('.js-like-count');
+
+    $likeBtn.blur();
+
+    if (liked === 'false') {
```
tlconnor: If you hadn't converted these to strings this would be cleaner:

```
if(liked) {
  ...
} else {
  ...
}
```

tlconnor: Also its generally nicer to put the positive case first in an if/else block.

app/controllers/images_controller.rb
tlconnor: Actually its not clear what the meaning of `liked` in this response is.  Does it mean the user used to like the image, or does it mean the user now likes the image?

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,44 @@
...
+      $likeCountField.text(`${++likeCount}`);
```
tlconnor: If you passed the new `likeCount` from the server instead of the old count you wouldn't need to do the ++/-- here.

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,44 @@
+import setFlashMessage from './set_flash_message'
+
+class LikeButton {
+  constructor(likeBtnClass) {
+    this.likeBtnClass = likeBtnClass;
+  }
+
+  updateLikeBtn({likeCount, liked}, $likeBtn) {
+    var $likeCountField = $likeBtn.find('.js-like-count');
```
tlconnor: In es6 you should be using `const` / `let` instead of `var`

app/assets/javascripts/images/like_button_driver.js
```diff
@@ -0,0 +1,5 @@
+import LikeButton from './like_button';
+
+const likeButton = new LikeButton('.js-like-btn');
```
tlconnor: I wonder if instead of passing a css class you could just pass the DOM element of the button?  That way you wouldn't need to keep looking it up.

Gemfile
```diff
@@ -31,6 +31,7 @@ end
 gem 'validate_url'
 gem 'email_validator'
 gem 'pundit'
+gem 'font-awesome-rails'
```
tlconnor: This doesn't seem to be working correctly.  The icon on the like button is showing up as a square.

https://ragurneys-pr-57.herokuapp.com/

app/assets/stylesheets/_image-card.scss
```diff
@@ -17,3 +17,12 @@
   font-size: 1em;
   margin-bottom: 2px;
 }
+
+.image-card__liked-false {
```
tlconnor: You shouldn't need to repeat the `float:right` here.  It would be better to instead use two CSS classes:

```
.image-card__liked {
  float: right;

  &.image-card__liked-true {
    background-color: #c1d9ff;
  }
}
```

app/views/images/_image_card.html.erb
```diff
@@ -23,5 +23,10 @@
                   class: "js-delete-image image-card__btn btn btn-danger",
                   'data-confirm': 'Are you sure you want to delete this image?' if policy(image).destroy? %>
     </div>
+    <%= render partial: 'like_button',
```
tlconnor: When rendering partials in views you don't actually need the `partial:` and `locals:` keywords.

```
<%= render 'like_button', { 
  image: image,
  liked: image_liked_by_current_user?(image)
} %>
```

test/models/like_test.rb
ragurney: 070e9b2

test/models/like_test.rb
ragurney: 070e9b2

test/models/like_test.rb
ragurney: 070e9b2

app/controllers/images_controller.rb
ragurney: This has been removed in this commit: de202f9

db/migrate/20161028173540_add_likes_table_and_add_like_to_image_and_user.rb
tlconnor: 👍

test/models/like_test.rb
tlconnor: 👍

test/controllers/images_controller_test.rb
tlconnor: 👍

app/controllers/images_controller.rb
tlconnor: I'll comment further when reviewing de202f9

tlconnor:
I notice the app isn't building properly anymore.  Could you please have a look? https://ragurneys-pr-57.herokuapp.com/

ragurney:
@tlconnor I'm working on fixing it right now

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,50 @@
+import setFlashMessage from './set_flash_message'
+
+class LikeButton {
+  constructor(element) {
+    this.element = element;
+    this.attachEventHandlers();
+    this.csrfToken = $('meta[name="csrf-token"]').attr('content');
```
tlconnor: Is there a way you could do this without jquery?  Perhaps using `querySelector`?

https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  updateLikeBtn({likeCount}) {
+    let likeCountField = this.element.getElementsByClassName('js-like-count')[0];
```
tlconnor: Generally in ES6 if you have a value that is not going to change you should use `const`.  This does two things:
1. it ensures the value is never reassigned
2. it makes the code more readable

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,50 @@
...
+
+    this.element.blur();
+    likeCountField.innerHTML = `${likeCount}`;
```
tlconnor: Is string interpolation required here, or is there a simpler way?

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,50 @@
...
+    this.element.classList.toggle('image-card__liked-true')
```
tlconnor: I like that there is now just one css class used when switching between the true/false states.  I wonder if it is even necessary to include the `-true` in the class name now.

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,50 @@
...
+        }
+      })
+        .then(response => {
+          if(response.status == 404) {
+            window.location.replace('/');
+          }
+          else if(response.status == 401) {
+            setFlashMessage('danger', 'You must log in to like images!');
+          }
+          else {
```
tlconnor: When we made these changes yesterday I mentioned one of the key differences between `fetch` and `$.get`: fetch does not raise errors on `non-200` responses.

The code here is handling the `404` and `401` cases, but any other unexpected responses would now be treated as success.

The only truly successful response codes are in the `2xx` range, so ideally this would be updated to parse as json for `2xx` and alert for anything else.  The alert could be achieved by raising an Error, which would then be caught at the end of the promise chain.

app/assets/javascripts/images/like_button.js
```diff
@@ -0,0 +1,50 @@
...
+            return response.json();
+          }
+        })
+        .then( data => {
+          if (data != null) {
```
tlconnor: Is there a simpler way to write the `if` block?

```
if (data != null) {
 ...
}
```

app/assets/stylesheets/_image-card.scss
```diff
@@ -17,3 +17,13 @@
   font-size: 1em;
   margin-bottom: 2px;
 }
+
+.image-card__liked {
```
tlconnor: I read this class name to mean that the image card has already been liked.  Maybe something like `image-card__likebutton` would be more suitable

app/assets/stylesheets/_image-card.scss
```diff
@@ -17,3 +17,13 @@
...
+  float: right;
+
+  &.image-card__liked-true {
```
tlconnor: Then if you make the change above the `-true` would become superfluous.

app/helpers/like_button_helper.rb
```diff
@@ -0,0 +1,5 @@
+module LikeButtonHelper
+  def image_liked_by_current_user?(image)
+    Like.find_by(image: image, user: current_user).present?
```
tlconnor: This line works, but is not the most efficient.  It is loading the record from the DB, initializing the model instance, then `.present?`checks if it exists.  If you run the line in the console you'll see a query that looks something like:

```
Like Load (0.4ms)  SELECT  `likes`.* FROM `likes` WHERE `likes`.`image_id` = 1 AND `likes`.`user_id` = 1 LIMIT 1
```

ActiveRecord already provides ways of checking if records exist that result in a faster query and less work within ruby.  The query run would then be

```
Like Exists (0.1ms)  SELECT 1 AS one FROM `likes` WHERE `likes`.`image_id` = 1 AND `likes`.`user_id` = 1 LIMIT 1
```

with the result being a simple `true`/`false` rather than an ActiveRecord object.

app/controllers/images_controller.rb
```diff
@@ -57,6 +57,16 @@ def destroy
     redirect_to images_path, flash: { success: 'Image successfully deleted!' }
   end
 
+  def like
+    authorize @image
+
+    liked = Like.find_by(image: @image, user: current_user)
```
tlconnor: In the review of ccb7ed9 I noted a better way of determining if a record exists.  I notice here that there are multiple places in the code that check if a `Like` record exists for a given image/user combination.  This could be `DRY`'d by adding something like:

```
@image.liked_by?(current_user)
```

or alternatively

```
current_user.likes?(@image)
```


If you do this there is also no need for `LikeButtonHelper` anymore.

app/controllers/images_controller.rb
```diff
@@ -104,6 +114,10 @@ def respond_to_invalid_share_form
   end
 
   def user_not_authorized
-    redirect_to new_session_path, flash: { danger: 'You must log in before accessing that page!' }
+    if request.content_type == 'application/json'
```
tlconnor: Controllers already provide a way to do different behavior for different content types.

I would suggest looking into http://apidock.com/rails/ActionController/MimeResponds/InstanceMethods/respond_to

app/assets/javascripts/images/like_button.js
ragurney: 3e1c85e

app/assets/javascripts/images/like_button.js
ragurney: 6c32cfd

app/assets/javascripts/images/like_button.js
ragurney: f47a3c8, 22d9b90

app/assets/stylesheets/_image-card.scss
ragurney: f47a3c8, 22d9b90

app/assets/stylesheets/_image-card.scss
ragurney: f47a3c8, 22d9b90

app/assets/javascripts/images/like_button.js
ragurney: 8cb73c4

app/assets/javascripts/images/like_button.js
ragurney: 69791b3

app/helpers/like_button_helper.rb
ragurney: 599422b, ce8cff7,

app/controllers/images_controller.rb
ragurney: 0c9e06d, 4ed3a3f

ragurney:
Just pushed and linked fixes addressing your comments.

app/models/image.rb
```diff
@@ -2,7 +2,13 @@
...
 class Image < ActiveRecord::Base
   belongs_to :user
+  has_many :likes, dependent: :destroy
   acts_as_taggable
+
   validates :url, presence: true, url: { allow_blank: true, message: 'must be a valid url' }
   validates :tag_list, presence: { message: 'must have at least one tag' }
+
+  def liked_by?(current_user)
+    current_user.nil? ? false : Like.exists?(image: self, user: current_user)
```
tlconnor: Is the `nil?` check for current_user really necessary?  If the user was `nil` I would expect the `.exists?` to return false.

Alternatively you could do something like

```
current_user && Like.exists?(image: self, user: current_user)
```

I don't expect you to change this, just FYI.

ragurney:
^ @kkajla12?

